PR #34, opened 2014-12-29
WIP: Adding location to metrics

goizueta:
@Kurt-von-Laven I'm running into some trouble testing this. Would you mind taking a preliminary look when you get back from break and let me know if I'm on the right track? Thanks.


Kurt-von-Laven:
@goizueta any concerns with closing this pull request and deleting this branch now that Philip added the location metric using GeoClue?

goizueta:
None at all. Let's give her the axe.

PR #38, opened 2015-03-20
Record location metrics

ptomato:
[endlessm/eos-sdk#861]

### DIFF
--- a/src/eins-location.c
+++ b/src/eins-location.c
@@ -1,0 +1,197 @@
+/* Copyright 2014, 2015 Endless Mobile, Inc. */
+
+/* This file is part of eos-metrics-instrumentation.
+ *
+ * eos-metrics-instrumentation is free software: you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 2 of the License, or (at your
+ * option) any later version.
+ *
+ * eos-metrics-instrumentation is distributed in the hope that it will be
+ * useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
+ * Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with eos-metrics-instrumentation.  If not, see
+ * <http://www.gnu.org/licenses/>.
+ */
+
+#include <glib.h>
+#include <eosmetrics/eosmetrics.h>
+
+#include "eins-location.h"
+#include "geoclue.h"
+
+#define DESKTOP_ID "eos-metrics-instrumentation"
+
+/* This is from geoclue private header src/public-api/gclue-enums.h */
+#define GCLUE_ACCURACY_LEVEL_CITY 4
+/* This should be kept in sync with geocode-glib/geocode-glib.h */
+#define LOCATION_ACCURACY_CITY 15000 /* m */
+
+/*
+ * Recorded once per boot. The payload contains the following information:
+ *   latitude (double)
+ *   longitude (double)
+ *   is altitude known? (boolean)
+ *   altitude (double) -- garbage if altitude is not known
+ *   accuracy of this location (double)
+ */
+#define EVENT_USER_LOCATION "abe7af92-6704-4d34-93cf-8f1b46eb09b8"
+
+static void
+on_location_proxy_ready (GObject      *obj,
+                         GAsyncResult *res)
+{
+  GError *error = NULL;
+
+  GeoclueLocation *location = geoclue_location_proxy_new_for_bus_finish (res,
+                                                                         &error);
+  if (error != NULL)
+    {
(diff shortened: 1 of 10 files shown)


src/eins-location.c
```diff
@@ -0,0 +1,197 @@
...
+      if (!g_error_matches (error, G_IO_ERROR, G_IO_ERROR_CANCELLED))
+        g_critical ("Failed to connect to GeoClue2 service: %s",
+                    error->message);
+      g_error_free (error);
+      return;
+    }
+
+  geoclue_manager_call_get_client (manager, NULL,  /* cancellable */
+                                   (GAsyncReadyCallback) on_get_client_ready,
+                                   NULL /* data */);
```
cosimoc: I think you could unref the manager here

src/eins-location.c
```diff
@@ -0,0 +1,197 @@
...
+                     GAsyncResult   *res)
+{
+  gchar *client_path;
+  GError *error = NULL;
+
+  if (!geoclue_manager_call_get_client_finish (manager, &client_path, res,
+                                               &error))
+    {
+      if (!g_error_matches (error, G_IO_ERROR, G_IO_ERROR_CANCELLED))
+        g_critical ("Failed to connect to GeoClue2 service: %s",
```
cosimoc: I would use a more specific message here mentioning we failed to get client from manager

src/eins-location.c
```diff
@@ -0,0 +1,197 @@
...
+on_client_proxy_ready (GObject      *obj,
+                       GAsyncResult *res)
+{
+  GError *error = NULL;
+
+  GeoclueClient *client = geoclue_client_proxy_new_for_bus_finish (res, &error);
+  if (error != NULL)
+    {
+      if (!g_error_matches (error, G_IO_ERROR, G_IO_ERROR_CANCELLED))
+        g_critical ("Failed to connect to GeoClue2 service: %s",
```
cosimoc: Again I would use a more specific message that says we couldn't create a client proxy

src/eins-location.c
```diff
@@ -0,0 +1,197 @@
...
+static void
+on_start_ready (GeoclueClient *client,
+                GAsyncResult  *res)
+{
+  GError *error = NULL;
+
+  if (!geoclue_client_call_start_finish (client, res, &error))
+    {
+      if (!g_error_matches (error, G_IO_ERROR, G_IO_ERROR_CANCELLED))
+        g_critical ("Failed to start GeoClue2 client: %s", error->message);
```
cosimoc: Also here I would use a more specific message

src/eins-location.c
```diff
@@ -0,0 +1,197 @@
...
+                         GAsyncResult *res)
+{
+  GError *error = NULL;
+
+  GeoclueLocation *location = geoclue_location_proxy_new_for_bus_finish (res,
+                                                                         &error);
+  if (error != NULL)
+    {
+      if (!g_error_matches (error, G_IO_ERROR, G_IO_ERROR_CANCELLED))
+        g_critical ("Failed to connect to GeoClue2 service: %s",
```
cosimoc: Ditto about error messages

src/eins-location.c
```diff
@@ -0,0 +1,197 @@
...
+    }
+}
+
+static void
+on_client_proxy_ready (GObject      *obj,
+                       GAsyncResult *res)
+{
+  GError *error = NULL;
+
+  GeoclueClient *client = geoclue_client_proxy_new_for_bus_finish (res, &error);
```
cosimoc: Nothing unrefs the client. I don't know that we need to record location updates as part of this issue, so perhaps you can just unref it after calling `call_start()`, if the `location-updated` signal is not emitted after the callback.
Otherwise it's fine also to just ignore this leak, since it's a singleton object.

tests/test-location.py
```diff
@@ -0,0 +1,210 @@
...
+            METRICS_IFACE, system_bus=True)
+        self.metrics_mock = dbus.Interface(self.dbus_con.get_object(
+            METRICS_OBJECT, METRICS_PATH), dbusmock.MOCK_IFACE)
+        self.metrics_mock.AddMethod('', 'RecordSingularEvent', 'uayxbv', '', '')
+
+        # Mechanism for blocking on a particular call
+        self.dbus_con.add_signal_receiver(self.handle_dbus_event_received,
+                                          signal_name='MethodCalled')
+        self.mainloop = GLib.MainLoop()
+        self._quit_on = ''
```
cosimoc: Should be `self._quit_on_method` I think

src/eins-location.c
ptomato: I think we could unref the client after receiving the `location-updated` signal but then we wouldn't get further updates if the user moved their computer far enough above the threshold of 15 km. That's very unlikely anyway, so I'll go ahead and unref it.

src/eins-location.c
ptomato: I can't think of a more specific message for this one - literally, the client object's `Start` method failed.

src/eins-location.c
ptomato: Fixed

PR #40, opened 2015-03-23
Launch after event recorder daemon, and flush the system DBus connection on shutdown.

Kurt-von-Laven:
Launch after event recorder daemon.

Since the metrics instrumentation daemon requires the event recorder
daemon in order to record any events, we make it list the event
recorder daemon as a dependency and ensure that the metrics
instrumentation daemon launches after and closes before the event
recorder daemon. This also helps the shutdown event have enough time to
be persisted to disk by the event recorder daemon since it is recorded
when the instrumentation daemon receives the SIGTERM signal from
systemd.

Flush the system DBus connection on shutdown.

This is important for ensuring that the shutdown event makes it to the
event recorder daemon in time to be flushed to disk before power is cut.

[endlessm/eos-sdk#2906]

### DIFF
--- a/data/eos-metrics-instrumentation.service.in
+++ b/data/eos-metrics-instrumentation.service.in
@@ -1,8 +1,8 @@
 [Unit]
 Description=EndlessOS Metrics Instrumentation
-Requires=dbus.service
+Requires=dbus.service eos-metrics-event-recorder.service
 Wants=dbus-org.freedesktop.GeoClue2.service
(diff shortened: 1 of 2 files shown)


ptomato:
Needs a rebase.

Kurt-von-Laven:
Rebased!

cosimoc:
@Kurt-von-Laven This is Revise ready

Kurt-von-Laven:
@cosimoc CR ready!

PR #42, opened 2015-04-08
Depend on version 0.3.0+ of event recorder API.

Kurt-von-Laven:
New synchronous methods were added to the event recorder API in version
0.3.0. These methods are required in the instrumentation daemon for
robust recording of logout and shutdown.

[endlessm/eos-sdk#3029]

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -4,7 +4,7 @@ Priority: standard
 Maintainer: Kurt von Laven <kurt@endlessm.com>
 Build-Depends: debhelper (>= 9),
 	autotools-dev,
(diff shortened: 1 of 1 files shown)


debian/control
```diff
@@ -4,7 +4,7 @@ Priority: standard
 Maintainer: Kurt von Laven <kurt@endlessm.com>
 Build-Depends: debhelper (>= 9),
 	autotools-dev,
-	eos-metrics-0-dev,
+	eos-metrics-0-dev (>= 0.4),
```
ptomato: Should be 0.3.0

PR #43, opened 2015-04-08
Use synchronous event recorder API for shutdown and logouts.

Kurt-von-Laven:
We currently flush outgoing messages at shutdown but don't wait for the
event recorder daemon to handle those method calls before exiting. This
creates a race condition where systemd can shutdown the event recorder
daemon before it has a chance to handle the pending method calls. We now
instead call the event recorder daemon synchronously when recording the
shutdown event. Since we instruct systemd to shutdown the event recorder
daemon after the instrumentation daemon, this ensures the shutdown event
is not dropped.

We don't need to inhibit shutdown anymore because we can record logout
events synchronously. Treating SIGTERM as a logout of any remaining
users is more robust than treating the PrepareForShutdown signal this
way as it's far less likely that shutdown will be cancelled after the
SIGTERM.

[endlessm/eos-sdk#3029]

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -53,6 +53,8 @@ AC_CONFIG_SRCDIR([src/eos-metrics-instrumentation.c])
 # no-portability suppresses warnings about syntax specific to GNU make
 # parallel-tests specifies that we use the new parallel-running test harness.
 # Unlike serial-tests, this option is accepted by Automake 1.11
+# subdir-objects ensures forward compatibility with Automake 2.0 and can be
+# removed when this behavior becomes the default.
 # tar-ustar is required because there may be files whose entire paths exceed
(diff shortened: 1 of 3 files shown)


configure.ac
```diff
@@ -80,7 +81,7 @@ AC_DEFINE([EOS_INSTRUMENTATION_VERSION], [_EOS_INSTRUMENTATION_VERSION_MACRO], [
 
 # Required versions of libraries
 # Update these whenever you use a function that requires a certain API version
-EOS_METRICS_REQUIREMENT="eosmetrics-0"
+EOS_METRICS_REQUIREMENT="eosmetrics-0 >= 0.4"
```
ptomato: Should be 0.3.0

PR #44, opened 2015-05-02
Upgrade from Python 2 to Python 3.

Kurt-von-Laven:
Add Python m4 modules, and require python3-dbusmock 0.10+.

[endlessm/eos-sdk#3038]

### DIFF
--- a/tests/test-location.py
+++ b/tests/test-location.py
@@ -181,8 +181,10 @@ class TestLocationIntegration(dbusmock.DBusTestCase):
         self.quit_on('RecordSingularEvent')
         self.mainloop.run()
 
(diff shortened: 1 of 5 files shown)


tests/test-location.py
```diff
@@ -22,6 +22,7 @@
 import subprocess
 import tempfile
 import unittest
+import uuid
```
ptomato: If you're using `uuid` anyway, can you convert the event ID to a UUID object directly, and use whatever comparison facility the UUID library gives you?

Kurt-von-Laven: Yep, sounds good to me. (`assertEquals` appears to work correctly with UUIDs.)

ptomato:
**CR pass**

PR #50, opened 2015-08-11
Make test-location.py executable and add shebang.

Kurt-von-Laven:
[endlessm/eos-sdk#3151]


Kurt-von-Laven:
This PR is not intended to be cherry-picked to eos2.4 but can be for the sake of convenience.

ptomato:
**CR pass**

PR #52, opened 2015-08-14
Sdk/3477

magcius:
[endlessm/eos-sdk#3477]

### DIFF
--- a/src/eos-metrics-instrumentation.c
+++ b/src/eos-metrics-instrumentation.c
@@ -217,17 +217,16 @@ systemd_dbus_proxy_new (void)
 }
 
 /*
- * Populate user_id with the user ID of the user associated with the SessionNew
- * or SessionRemoved signal with the given parameters. user_id must already be
- * allocated and non-NULL. Return TRUE if user_id was successfully populated and
+ * Populate user_id with the user ID of the user associated with the given
+ * logind session object at session_path. user_id must already be allocated
+ * and non-NULL. Return TRUE if user_id was successfully populated and
  * FALSE otherwise, in which case its contents should be ignored.
  */
 static gboolean
-get_user_id (GVariant *session_parameters, guint32 *user_id)
+get_user_id (const gchar *session_path,
+             guint32     *user_id)
 {
     GError *error = NULL;
-    gchar *session_path;
-    g_variant_get_child (session_parameters, 1, "o", &session_path);
     GDBusProxy *dbus_proxy =
       g_dbus_proxy_new_for_bus_sync (G_BUS_TYPE_SYSTEM,
                                      G_DBUS_PROXY_FLAGS_NONE,
@@ -237,7 +236,6 @@ get_user_id (GVariant *session_parameters, guint32 *user_id)
                                      "org.freedesktop.DBus.Properties",
                                      NULL /* GCancellable */,
                                      &error);
-    g_free (session_path);
 
     if (dbus_proxy == NULL)
       {
@@ -273,57 +271,51 @@ get_user_id (GVariant *session_parameters, guint32 *user_id)
     return TRUE;
 }
 
-/*
- * Return TRUE if the given parameters of a SessionNew or SessionRemoved signal
- * correspond to a human session. If the given parameters don't correspond to a
- * human session, or this can't be determined, return FALSE.
- */
 static gboolean
-is_human_session (GVariant *session_parameters)
+is_human_session (const gchar *session_id)
 {
-    guint32 user_id;
-    if (!get_user_id (session_parameters, &user_id))
-      return FALSE;
-    return user_id >= MIN_HUMAN_USER_ID;
+    /* All normal user sessions start with digits -- greeter sessions
+     * start with 'c'.
+     */
+    return g_ascii_isdigit (session_id[0]);
+}
+
+static gboolean
+session_in_set (const gchar *session_id)
+{
+    gpointer data = g_datalist_get_data (&humanity_by_session_id, session_id);
+    return (gboolean) GPOINTER_TO_UINT (data);
 }
 
 /*
- * If the given parameters of a SessionNew or SessionRemoved signal correspond
- * to a human session, return TRUE and add the corresponding session_id to the
+ * If the given session_id corresponds to a human session not already in the
+ * set, return TRUE and add the corresponding session_id to the
  * humanity_by_session_id set. Otherwise, return FALSE.
  */
 static gboolean
-add_session_to_set (GVariant *session_parameters)
+add_session_to_set (const gchar *session_id)
 {
-    if (!is_human_session (session_parameters))
+    if (!is_human_session (session_id) || session_in_set (session_id))
       return FALSE;
 
-    gchar *session_id;
-    g_variant_get_child (session_parameters, 0, "s", &session_id);
-    g_datalist_set_data (&humanity_by_session_id, session_id, (gpointer) TRUE);
-    g_free (session_id);
+    g_datalist_set_data (&humanity_by_session_id, session_id,
+                         GUINT_TO_POINTER (TRUE));
     return TRUE;
 }
 
 /*
- * If the given parameters of a SessionNew or SessionRemoved signal correspond
- * to a human session that is currently in the humanity_by_session_id set,
- * return TRUE and remove the session from the set. Otherwise, return FALSE.
+ * If the given session_id corresponds to a human session tracked inside the
+ * humanity_by_session_id set, remove it and return TRUE. Otherwise, return
+ * FALSE.
  */
 static gboolean
-remove_session_from_set (GVariant *session_parameters)
+remove_session_from_set (const gchar *session_id)
 {
-    gchar *session_id;
-    g_variant_get_child (session_parameters, 0, "s", &session_id);
+    if (!is_human_session (session_id) || !session_in_set (session_id))
+      return FALSE;
 
-    gboolean is_human = (gboolean) g_datalist_get_data (&humanity_by_session_id,
-                                                        session_id);
-
-    if (is_human)
-      g_datalist_remove_data (&humanity_by_session_id, session_id);
-
-    g_free (session_id);
-    return is_human;
+    g_datalist_remove_data (&humanity_by_session_id, session_id);
+    return TRUE;
 }
 
 /*
@@ -340,6 +332,34 @@ record_stop_for_login (GQuark   session_id_quark,
     emtr_event_recorder_record_stop_sync (emtr_event_recorder_get_default (),
                                           USER_IS_LOGGED_IN,
                                           session_id_variant,
+                                          NULL /* auxiliary_payload */);
+}
+
+static void
+add_session (const gchar *session_id,
+             guint32      user_id)
+{
+    if (!add_session_to_set (session_id))
+      return;
+
+    GVariant *user_id_variant =
+      (user_id >= MIN_HUMAN_USER_ID) ? g_variant_new_uint32 (user_id) : NULL;
+
+    emtr_event_recorder_record_start (emtr_event_recorder_get_default (),
+                                      USER_IS_LOGGED_IN,
+                                      g_variant_new_string (session_id),
+                                      user_id_variant);
+}
+
+static void
+remove_session (const gchar *session_id)
+{
+    if (!remove_session_from_set (session_id))
+      return;
+
+    emtr_event_recorder_record_stop_sync (emtr_event_recorder_get_default (),
+                                          USER_IS_LOGGED_IN,
+                                          g_variant_new_string (session_id),
                                           NULL /* auxiliary_payload */);
 }
 
@@ -363,31 +383,23 @@ record_login (GDBusProxy *dbus_proxy,
               GVariant   *parameters,
               gpointer    user_data)
 {
-    if (strcmp ("SessionRemoved", signal_name) == 0 &&
-        remove_session_from_set (parameters))
+    if (strcmp ("SessionRemoved", signal_name) == 0)
       {
-        GVariant *session_id = g_variant_get_child_value (parameters, 0);
-        emtr_event_recorder_record_stop_sync (emtr_event_recorder_get_default (),
-                                              USER_IS_LOGGED_IN,
-                                              session_id,
-                                              NULL /* auxiliary_payload */);
-        g_variant_unref (session_id);
+        const gchar *session_id;
+        g_variant_get (parameters, "&s", &session_id);
+
+        remove_session (session_id);
       }
-    else if (strcmp ("SessionNew", signal_name) == 0 &&
-             add_session_to_set (parameters))
+    else if (strcmp ("SessionNew", signal_name) == 0)
       {
-        GVariant *session_id = g_variant_get_child_value (parameters, 0);
+        const gchar *session_id, *session_path;
+        g_variant_get (parameters, "&s&o", &session_id, &session_path);
 
         guint32 user_id;
-        gboolean user_id_is_valid = get_user_id (parameters, &user_id);
-        GVariant *user_id_variant = user_id_is_valid ?
-          g_variant_new_uint32 (user_id) : NULL;
+        if (!get_user_id (session_path, &user_id))
+          return;
 
-        emtr_event_recorder_record_start (emtr_event_recorder_get_default (),
-                                          USER_IS_LOGGED_IN,
-                                          session_id,
-                                          user_id_variant);
-        g_variant_unref (session_id);
+        add_session (session_id, user_id);
       }
 }
 
@@ -413,16 +425,40 @@ login_dbus_proxy_new (void)
                                      "org.freedesktop.login1.Manager",
                                      NULL /* GCancellable */,
                                      &error);
-    if (dbus_proxy == NULL)
+    if (error)
       {
-        g_warning ("Error creating GDBusProxy: %s\n", error->message);
+        g_warning ("Error creating GDBusProxy: %s.", error->message);
         g_error_free (error);
+        return NULL;
       }
-    else
+
+    g_signal_connect (dbus_proxy, "g-signal", G_CALLBACK (record_login),
+                      NULL /* data */);
+
+    GVariant *sessions =
+      g_dbus_proxy_call_sync (dbus_proxy, "ListSessions", NULL,
+                              G_DBUS_CALL_FLAGS_NONE, -1, NULL, &error);
+    if (error)
       {
-        g_signal_connect (dbus_proxy, "g-signal", G_CALLBACK (record_login),
-                          NULL /* data */);
+        g_warning ("Error calling ListSessions: %s.", error->message);
+        g_error_free (error);
+        return NULL;
       }
+
+    GVariantIter *session_iter;
+    g_variant_get (sessions, "a(susso)", &session_iter);
+
+    const gchar *session_id;
+    guint32 user_id;
+    while (g_variant_iter_loop (session_iter, "(&suss&o)", &session_id,
+                                &user_id, NULL, NULL, NULL))
+      {
+        add_session (session_id, user_id);
+      }
+
+    g_variant_iter_free (session_iter);
+    g_variant_unref (sessions);
+
     return dbus_proxy;
 }
 
--- a/src/eos-metrics-instrumentation.c
+++ b/src/eos-metrics-instrumentation.c
@@ -67,10 +67,7 @@
  */
 #define NETWORK_STATUS_CHANGED_EVENT "5fae6179-e108-4962-83be-c909259c0584"
 
-// Protected by humanity_by_session_id lock.
 static GData *humanity_by_session_id;
-
-G_LOCK_DEFINE_STATIC (humanity_by_session_id);
 
 static gboolean start_time_set = FALSE;
 
@@ -303,9 +300,7 @@ add_session_to_set (GVariant *session_parameters)
 
     gchar *session_id;
     g_variant_get_child (session_parameters, 0, "s", &session_id);
-    G_LOCK (humanity_by_session_id);
     g_datalist_set_data (&humanity_by_session_id, session_id, (gpointer) TRUE);
-    G_UNLOCK (humanity_by_session_id);
     g_free (session_id);
     return TRUE;
 }
@@ -321,14 +316,11 @@ remove_session_from_set (GVariant *session_parameters)
     gchar *session_id;
     g_variant_get_child (session_parameters, 0, "s", &session_id);
 
-    G_LOCK (humanity_by_session_id);
     gboolean is_human = (gboolean) g_datalist_get_data (&humanity_by_session_id,
                                                         session_id);
 
     if (is_human)
       g_datalist_remove_data (&humanity_by_session_id, session_id);
-
-    G_UNLOCK (humanity_by_session_id);
 
     g_free (session_id);
     return is_human;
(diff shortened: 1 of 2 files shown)


src/eos-metrics-instrumentation.c
```diff
@@ -296,17 +287,14 @@ is_human_session (GVariant *session_parameters)
...
 {
-    if (!is_human_session (session_parameters))
+    if (!is_human_session (session_id))
       return FALSE;
 
-    gchar *session_id;
-    g_variant_get_child (session_parameters, 0, "s", &session_id);
     G_LOCK (humanity_by_session_id);
-    g_datalist_set_data (&humanity_by_session_id, session_id, (gpointer) TRUE);
+    g_datalist_set_data (&humanity_by_session_id, session_id, GUINT_TO_POINTER (TRUE));
```
Kurt-von-Laven: For my own education, what is the rationale for using `GUINT_TO_POINTER` and `GPOINTER_TO_UINT`?

ptomato: Guaranteed to work even when `sizeof (gpointer) !== sizeof (int)`

Kurt-von-Laven: nit: Line exceeds 80 columns (multiple instances).

magcius: I wasn't aware that was part of our style. Will fix, thanks.

Kurt-von-Laven: Not sure if anyone already shared these with you:
* https://github.com/endlessm/eos-documentation/wiki/C-Coding-Style
* https://github.com/endlessm/eos-documentation/wiki/JavaScript-Coding-Style

src/eos-metrics-instrumentation.c
```diff
@@ -226,11 +226,9 @@ systemd_dbus_proxy_new (void)
  * FALSE otherwise, in which case its contents should be ignored.
  */
 static gboolean
-get_user_id (GVariant *session_parameters, guint32 *user_id)
+get_user_id (const char *session_path, guint32 *user_id)
```
Kurt-von-Laven: nit: Each parameter belongs on its own line (pre-existing).

src/eos-metrics-instrumentation.c
```diff
@@ -276,18 +273,12 @@ get_user_id (GVariant *session_parameters, guint32 *user_id)
...
 static gboolean
-is_human_session (GVariant *session_parameters)
+is_human_session (const char *session_id)
 {
-    guint32 user_id;
-    if (!get_user_id (session_parameters, &user_id))
-      return FALSE;
-    return user_id >= MIN_HUMAN_USER_ID;
+  /* All normal user sessions start with digits -- greeter sessions
+   * start with 'c'. */
```
Kurt-von-Laven: nit: Add extra blank line at end of comment.

```
/* foo
 * bar
 */
```

src/eos-metrics-instrumentation.c
```diff
@@ -351,6 +338,22 @@ record_stop_for_login (GQuark   session_id_quark,
                                           NULL /* auxiliary_payload */);
 }
 
+static void
+add_session (const char *session_id,
+             const char *session_path,
+             guint32 user_id)
+{
+  if (!add_session_to_set (session_id))
```
Kurt-von-Laven: nit: Recommend matching the existing indentation in this file even though it violates the style guide. At some point I will probably fix it all at once.

src/eos-metrics-instrumentation.c
Kurt-von-Laven: nit: `char` --> `gchar` for consistency (multiple instances)

src/eos-metrics-instrumentation.c
```diff
@@ -381,21 +384,18 @@ record_login (GDBusProxy *dbus_proxy,
...
       }
-    else if (strcmp ("SessionNew", signal_name) == 0 &&
-             add_session_to_set (parameters))
+    else if (strcmp ("SessionNew", signal_name) == 0)
       {
-        GVariant *session_id = g_variant_get_child_value (parameters, 0);
+        const char *session_id, *session_path;
+
+        g_variant_get (parameters, "&s&o",
+                       &session_id, &session_path);
```
Kurt-von-Laven: nit: fits on one line.

src/eos-metrics-instrumentation.c
```diff
@@ -381,21 +384,18 @@ record_login (GDBusProxy *dbus_proxy,
                                               NULL /* auxiliary_payload */);
         g_variant_unref (session_id);
       }
-    else if (strcmp ("SessionNew", signal_name) == 0 &&
-             add_session_to_set (parameters))
+    else if (strcmp ("SessionNew", signal_name) == 0)
       {
-        GVariant *session_id = g_variant_get_child_value (parameters, 0);
+        const char *session_id, *session_path;
+
```
Kurt-von-Laven: nit: rm blank line.

src/eos-metrics-instrumentation.c
```diff
@@ -413,27 +413,48 @@ record_logout_for_all_remaining_sessions (void)
...
+                                   G_DBUS_PROXY_FLAGS_NONE,
+                                   NULL /* GDBusInterfaceInfo */,
+                                   "org.freedesktop.login1",
+                                   "/org/freedesktop/login1",
+                                   "org.freedesktop.login1.Manager",
+                                   NULL /* GCancellable */,
+                                   &error);
+  if (error)
+    {
+      g_warning ("Error creating GDBusProxy: %s", error->message);
```
Kurt-von-Laven: nit: Missing terminal period.

src/eos-metrics-instrumentation.c
```diff
@@ -413,27 +413,48 @@ record_logout_for_all_remaining_sessions (void)
...
+                                               G_DBUS_CALL_FLAGS_NONE, -1,
+                                               NULL, &error);
+  if (error)
+    {
+      g_warning ("Error calling ListSessions: %s", error->message);
+      g_error_free (error);
+      return NULL;
+    }
+
+  GVariantIter *iter;
```
Kurt-von-Laven: nit: Prefer a more semantic variable name.

magcius: Feel free to suggest one.

src/eos-metrics-instrumentation.c
magcius: I've never seen a terminal period in any error messages.

src/eos-metrics-instrumentation.c
```diff
@@ -226,11 +226,9 @@ systemd_dbus_proxy_new (void)
  * FALSE otherwise, in which case its contents should be ignored.
  */
 static gboolean
-get_user_id (GVariant *session_parameters, guint32 *user_id)
+get_user_id (const char *session_path, guint32 *user_id)
 {
     GError *error = NULL;
-    gchar *session_path;
```
Kurt-von-Laven: This variable is still referenced in the call to `g_dbus_proxy_new_for_bus_sync` below.

src/eos-metrics-instrumentation.c
```diff
@@ -296,17 +287,14 @@ is_human_session (GVariant *session_parameters)
  * humanity_by_session_id set. Otherwise, return FALSE.
  */
 static gboolean
-add_session_to_set (GVariant *session_parameters)
+add_session_to_set (const char *session_id)
```
Kurt-von-Laven: nit: Recommend modifying `remove_session_from_set` accordingly.

src/eos-metrics-instrumentation.c
```diff
@@ -351,6 +338,22 @@ record_stop_for_login (GQuark   session_id_quark,
...
 
+static void
+add_session (const char *session_id,
+             const char *session_path,
+             guint32 user_id)
+{
+  if (!add_session_to_set (session_id))
+    return;
+
+  GVariant *user_id_variant = user_id > 0 ? g_variant_new_uint32 (user_id) : NULL;
```
Kurt-von-Laven: Seems like we should test for `user_id >= MIN_HUMAN_USER_ID` instead.

src/eos-metrics-instrumentation.c
```diff
@@ -351,6 +338,22 @@ record_stop_for_login (GQuark   session_id_quark,
                                           NULL /* auxiliary_payload */);
 }
 
+static void
+add_session (const char *session_id,
+             const char *session_path,
+             guint32 user_id)
```
Kurt-von-Laven: nit: `user_id` should be vertically aligned with `session_id` and `session_path`.

src/eos-metrics-instrumentation.c
```diff
@@ -226,11 +226,9 @@ systemd_dbus_proxy_new (void)
  * FALSE otherwise, in which case its contents should be ignored.
```
Kurt-von-Laven: nit: Comment needs updating.

src/eos-metrics-instrumentation.c
```diff
@@ -413,27 +413,48 @@ record_logout_for_all_remaining_sessions (void)
...
+  if (error)
+    {
+      g_warning ("Error creating GDBusProxy: %s", error->message);
+      g_error_free (error);
+      return NULL;
+    }
+
+  g_signal_connect (dbus_proxy, "g-signal", G_CALLBACK (record_login),
+                    NULL /* data */);
+  GVariant *sessions = g_dbus_proxy_call_sync (dbus_proxy, "ListSessions",
```
Kurt-von-Laven: nit: Recommend styling as follows.

```
GVariant *sessions =
  g_dbus_proxy_call_sync (dbus_proxy, "ListSessions", NULL,
                                            G_DBUS_CALL_FLAGS_NONE, -1, NULL, &error);
```

src/eos-metrics-instrumentation.c
Kurt-von-Laven: This (like most of my comments) is for consistency with the rest of the metrics system.

src/eos-metrics-instrumentation.c
Kurt-von-Laven: `session_iter`

src/eos-metrics-instrumentation.c
```diff
@@ -413,27 +413,48 @@ record_logout_for_all_remaining_sessions (void)
...
+                                   NULL /* GCancellable */,
+                                   &error);
+  if (error)
+    {
+      g_warning ("Error creating GDBusProxy: %s", error->message);
+      g_error_free (error);
+      return NULL;
+    }
+
+  g_signal_connect (dbus_proxy, "g-signal", G_CALLBACK (record_login),
```
Kurt-von-Laven: @cosimoc is there a race here? Are we atomically registering a signal handler for the `SessionNew` signal and calling `ListSessions`?

src/eos-metrics-instrumentation.c
magcius: I actually copied this message from another function entirely. Most messages don't have the period.

https://github.com/endlessm/eos-metrics-instrumentation/blob/master/src/eos-metrics-instrumentation.c#L247
https://github.com/endlessm/eos-metrics-instrumentation/blob/master/src/eos-metrics-instrumentation.c#L265
https://github.com/endlessm/eos-metrics-instrumentation/blob/master/src/eos-metrics-instrumentation.c#L428
https://github.com/endlessm/eos-metrics-instrumentation/blob/master/src/eos-metrics-instrumentation.c#L485

Kurt-von-Laven: Those newlines should all be periods; just haven't fixed those particular instances yet. Most messages in the metrics system definitely do end in periods.

src/eos-metrics-instrumentation.c
magcius: Are you asking if it's possible to have a new session appear in the time, and so it appears both in the ListSessions and in the callback? That's extremely unlikely, but even if so, we'll fizzle out because of the pre-existing humanity data.

Kurt-von-Laven: Good point. Forgot we were tracking things with a set.

src/eos-metrics-instrumentation.c
magcius: What needs updating here?

magcius: Oh, I missed that we were referring to `session_parameters`. Nevermind.

src/eos-metrics-instrumentation.c
Kurt-von-Laven: Ah, but we shouldn't call `emtr_event_recorder_start` twice in a row with the same key. I suppose we could modify `add_session_to_set` to refrain from calling `emtr_event_recorder_start` if the session ID is already in the set.

src/eos-metrics-instrumentation.c
```diff
@@ -70,8 +70,6 @@
 // Protected by humanity_by_session_id lock.
```
Kurt-von-Laven: rm

src/eos-metrics-instrumentation.c
```diff
@@ -276,62 +272,54 @@ get_user_id (GVariant *session_parameters, guint32 *user_id)
...
  */
 static gboolean
-add_session_to_set (GVariant *session_parameters)
+add_session_to_set (const gchar *session_id)
 {
-    if (!is_human_session (session_parameters))
+    if (!is_human_session (session_id))
+      return FALSE;
+
+    gboolean in_set = (gboolean) GPOINTER_TO_UINT (
```
Kurt-von-Laven: If you were to break the line here, our style guide would expect the following vertical alignment:

```
gboolean in_set = (gboolean) GPOINTER_TO_UINT (
+                                              g_datalist_get_data (&humanity_by_session_id, session_id));
```

Since that doesn't do anything to reduce the line length, I recommend splitting this into multiple statements instead.

magcius: That seems like it's a particularly silly style guide if it suggests a line break that doesn't help us meet the other rule about line length.

src/eos-metrics-instrumentation.c
```diff
@@ -276,62 +272,54 @@ get_user_id (GVariant *session_parameters, guint32 *user_id)
...
+     * start with 'c'.
+     */
+    return g_ascii_isdigit (session_id[0]);
 }
 
 /*
- * If the given parameters of a SessionNew or SessionRemoved signal correspond
- * to a human session, return TRUE and add the corresponding session_id to the
- * humanity_by_session_id set. Otherwise, return FALSE.
+ * If the given session_id corresponds to a human session not already in the set
```
Kurt-von-Laven: The comma before "return TRUE" should remain.

src/eos-metrics-instrumentation.c
Kurt-von-Laven: I don't understand; the style guide doesn't suggest this line break.

magcius: Well, you said "If you were to break the line here, our style guide would expect the following vertical alignment"

At which point it seems like a silly style guide if it recommends a line break to meet 80 chars, and then vertical alignment that would make it not meet 80 chars.

Anyway, I fixed it in a different way.

src/eos-metrics-instrumentation.c
```diff
@@ -276,62 +272,54 @@ get_user_id (GVariant *session_parameters, guint32 *user_id)
...
-    gchar *session_id;
-    g_variant_get_child (session_parameters, 0, "s", &session_id);
-
-    G_LOCK (humanity_by_session_id);
-    gboolean is_human = (gboolean) g_datalist_get_data (&humanity_by_session_id,
-                                                        session_id);
-
-    if (is_human)
-      g_datalist_remove_data (&humanity_by_session_id, session_id);
+    if (!is_human_session (session_id))
```
Kurt-von-Laven: I believe we can omit this test.

src/eos-metrics-instrumentation.c
```diff
@@ -276,62 +272,54 @@ get_user_id (GVariant *session_parameters, guint32 *user_id)
...
+      return FALSE;
 
-    G_UNLOCK (humanity_by_session_id);
+    gboolean in_set = (gboolean) GPOINTER_TO_UINT (
```
Kurt-von-Laven: nit: Consider decomposing this pattern since it appears twice now.

src/eos-metrics-instrumentation.c
magcius: We can, but I think the parallels of the early-return make a bit of sense.

src/eos-metrics-instrumentation.c
```diff
@@ -351,6 +339,34 @@ record_stop_for_login (GQuark   session_id_quark,
...
 
+static void
+add_session (const gchar *session_id,
+             guint32      user_id)
+{
+    if (!add_session_to_set (session_id))
+      return;
+
+    GVariant *user_id_variant = ((user_id >= MIN_HUMAN_USER_ID) ?
+                                 g_variant_new_uint32 (user_id) : NULL);
```
Kurt-von-Laven: nit: Consider the following for readability:

```
GVariant *user_id_variant =
  ((user_id >= MIN_HUMAN_USER_ID) ? g_variant_new_uint32 (user_id) : NULL);
```

src/eos-metrics-instrumentation.c
```diff
@@ -351,6 +339,34 @@ record_stop_for_login (GQuark   session_id_quark,
...
 }
 
+static void
+add_session (const gchar *session_id,
+             guint32      user_id)
+{
+    if (!add_session_to_set (session_id))
+      return;
+
+    GVariant *user_id_variant = ((user_id >= MIN_HUMAN_USER_ID) ?
```
Kurt-von-Laven: nit: rm Outermost parentheses.

src/eos-metrics-instrumentation.c
magcius: Subjective -- I actually prefer my version, but I'll change it since you're the maintainer.

src/eos-metrics-instrumentation.c
magcius: This is for indentation purposes -- otherwise indentation tools won't align it with the expression below.

Kurt-von-Laven: Ah, I see.

src/eos-metrics-instrumentation.c
Kurt-von-Laven: Fine by me.

src/eos-metrics-instrumentation.c
```diff
@@ -371,43 +387,33 @@ record_login (GDBusProxy *dbus_proxy,
...
-        remove_session_from_set (parameters))
+    if (strcmp ("SessionRemoved", signal_name) == 0)
       {
-        GVariant *session_id = g_variant_get_child_value (parameters, 0);
-        emtr_event_recorder_record_stop_sync (emtr_event_recorder_get_default (),
-                                              USER_IS_LOGGED_IN,
-                                              session_id,
-                                              NULL /* auxiliary_payload */);
-        g_variant_unref (session_id);
+        const gchar *session_id;
```
Kurt-von-Laven: This is leaked AFAICS.

Kurt-von-Laven: `session_id` shouldn't be `const`.

magcius: It's not. The `&s` in the typestring returns a reference to inside the variant.

src/eos-metrics-instrumentation.c
```diff
@@ -371,43 +387,33 @@ record_login (GDBusProxy *dbus_proxy,
...
+        g_variant_get (parameters, "&s", &session_id);
+
+        remove_session (session_id);
       }
-    else if (strcmp ("SessionNew", signal_name) == 0 &&
-             add_session_to_set (parameters))
+    else if (strcmp ("SessionNew", signal_name) == 0)
       {
-        GVariant *session_id = g_variant_get_child_value (parameters, 0);
+        const gchar *session_id, *session_path;
```
Kurt-von-Laven: These are both leaked AFAICS and shouldn't be `const`.

src/eos-metrics-instrumentation.c
magcius: See https://developer.gnome.org/glib/unstable/gvariant-format-strings.html

src/eos-metrics-instrumentation.c
```diff
@@ -423,23 +429,44 @@ login_dbus_proxy_new (void)
...
         g_error_free (error);
+        return NULL;
       }
-    else
+
+    g_signal_connect (dbus_proxy, "g-signal", G_CALLBACK (record_login),
+                      NULL /* data */);
+
+    GVariant *sessions =
+      g_dbus_proxy_call_sync (dbus_proxy, "ListSessions",
```
Kurt-von-Laven: nit: Can squish this call onto fewer lines now that you broke the line after the assignment.

src/eos-metrics-instrumentation.c
Kurt-von-Laven: That's wicked cool.

magcius: Yeah, GVariant has a lot of weird features stuffed in it. I noticed you weren't writing very idiomatic GVariant with `get_child_value` and such, so I figured I'd take the time to clean it up.

src/eos-metrics-instrumentation.c
```diff
@@ -276,62 +271,57 @@ get_user_id (GVariant *session_parameters, guint32 *user_id)
...
+{
+    gpointer data = g_datalist_get_data (&humanity_by_session_id, session_id);
+    return (gboolean) GPOINTER_TO_UINT (data);
 }
 
 /*
- * If the given parameters of a SessionNew or SessionRemoved signal correspond
- * to a human session, return TRUE and add the corresponding session_id to the
- * humanity_by_session_id set. Otherwise, return FALSE.
+ * If the given session_id corresponds to a human session not already in the set,
```
Kurt-von-Laven: nit: These two lines exceed 80 columns.

src/eos-metrics-instrumentation.c
```diff
@@ -423,23 +431,44 @@ login_dbus_proxy_new (void)
...
+        g_error_free (error);
+        return NULL;
       }
+
+    GVariantIter *session_iter;
+    g_variant_get (sessions, "a(susso)", &session_iter);
+
+    const gchar *session_id;
+    guint32 user_id;
+    while (g_variant_iter_loop (session_iter, "(&suss&o)", &session_id, &user_id, NULL, NULL, NULL))
```
Kurt-von-Laven: nit: Line exceeds 80 columns.

src/eos-metrics-instrumentation.c
```diff
@@ -276,62 +271,57 @@ get_user_id (GVariant *session_parameters, guint32 *user_id)
...
-    gboolean is_human = (gboolean) g_datalist_get_data (&humanity_by_session_id,
-                                                        session_id);
-
-    if (is_human)
-      g_datalist_remove_data (&humanity_by_session_id, session_id);
+    if (!is_human_session (session_id))
+      return FALSE;
 
-    G_UNLOCK (humanity_by_session_id);
+    if (!session_in_set (session_id))
```
Kurt-von-Laven: nit: Can or this with the test above.

src/eos-metrics-instrumentation.c
```diff
@@ -276,62 +271,57 @@ get_user_id (GVariant *session_parameters, guint32 *user_id)
...
  */
 static gboolean
-add_session_to_set (GVariant *session_parameters)
+add_session_to_set (const gchar *session_id)
 {
-    if (!is_human_session (session_parameters))
+    if (!is_human_session (session_id))
+      return FALSE;
+
+    if (session_in_set (session_id))
```
Kurt-von-Laven: nit: Can or this with the test above.

Kurt-von-Laven:
**CR Pass**

PR #54, opened 2015-08-25
Fix syntax for g_variant_get() call

cosimoc:
We have a tuple here, so we should pass the right type, or we'll hit an
assertion.

[endlessm/eos-sdk#3477]


Kurt-von-Laven:
**CR Pass**

PR #55, opened 2015-08-27
test-location: don't assume first events are location

cosimoc:
This test currently assumes the first events recorded by the daemon will
be the location ones. This is not true anymore now that we also record
OS name, version and personality.
Change the code to wait for the specific event UUID instead.

[endlessm/eos-shell#5585]

### DIFF
--- a/tests/test-location.py
+++ b/tests/test-location.py
@@ -152,7 +152,7 @@ class TestLocationIntegration(dbusmock.DBusTestCase):
         Timeout after waiting for 20 seconds. Use like this:
             self.quit_on('MyMethod')
             self.mainloop.run()
-            # now MyMethod has been called
+            # Now MyMethod has been called.
         """
         self._quit_on_method = method_name
         GLib.timeout_add_seconds(20, self.fail, 'Test timed out after ' +
--- a/tests/test-location.py
+++ b/tests/test-location.py
@@ -128,6 +128,8 @@ class TestLocationIntegration(dbusmock.DBusTestCase):
                                           dbus_interface=dbusmock.MOCK_IFACE)
         self.mainloop = GLib.MainLoop()
         self._quit_on_method = ''
+        self._quit_on_uuid = None
+        self._event_args = None
 
         self.daemon = subprocess.Popen('./eos-metrics-instrumentation')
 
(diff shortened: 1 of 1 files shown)


tests/test-location.py
```diff
@@ -156,11 +158,34 @@ def quit_on(self, method_name):
         GLib.timeout_add_seconds(20, self.fail, 'Test timed out after ' +
                                  'waiting 20 seconds for D-Bus method call.')
 
+    def quit_on_event(self, event_uuid):
+        """Quit the main loop when RecordSingularEvent is called for @event_uuid.
+        Timeout after waiting for 20 seconds. Use like this:
+            self.quit_on_event(uuid.UUID('my-uuid'))
+            self.mainloop.run()
+            # now event my-uuid has been received
```
Kurt-von-Laven: nit: "now" --> "Now"

Kurt-von-Laven: nit: Terminal period.

tests/test-location.py
```diff
@@ -156,11 +158,34 @@ def quit_on(self, method_name):
...
+        """
+        self._quit_on_uuid = event_uuid
+        self.quit_on('RecordSingularEvent');
+
     def handle_dbus_event_received(self, name, *args):
         if name == self._quit_on_method:
+            if (self._quit_on_uuid):
```
Kurt-von-Laven: nit: rm parentheses for consistency with rest of Python style.

Kurt-von-Laven: I'm not sure what this does, so I usually use `if self._quit_on_uuid is not None`.

tests/test-location.py
```diff
@@ -156,11 +158,34 @@ def quit_on(self, method_name):
...
+                event_id_as_dbus_bytes = args[0][1]
+                event_id = self.dbus_bytes_to_uuid(event_id_as_dbus_bytes)
+                if (event_id != self._quit_on_uuid):
```
Kurt-von-Laven: nit: rm parentheses.

tests/test-location.py
```diff
@@ -156,11 +158,34 @@ def quit_on(self, method_name):
         GLib.timeout_add_seconds(20, self.fail, 'Test timed out after ' +
                                  'waiting 20 seconds for D-Bus method call.')
 
+    def quit_on_event(self, event_uuid):
```
Kurt-von-Laven: nit: Consider passing 'RecordSingularEvent' as a parameter or renaming this method to something like `quit_on_singular` since there are also aggregates and sequences.

tests/test-location.py
cosimoc: I'm following the style of `quit_on()`. I will change that too

Kurt-von-Laven:
**CR Pass**

PR #62, opened 2016-11-01
Log a metric when booted from live USB

wjt:
This will allow us to distinguish weird behaviour reported from
transient live boots, such as each boot appearing to be a fresh
installation.

https://phabricator.endlessm.com/T13890

### DIFF
--- a/src/eos-metrics-instrumentation.c
+++ b/src/eos-metrics-instrumentation.c
@@ -92,6 +92,18 @@
 #define PERSONALITY_CONFIG_GROUP "Personality"
 #define PERSONALITY_KEY "PersonalityName"
 
+/*
+ * Recorded once at startup when booted from a combined live + installer USB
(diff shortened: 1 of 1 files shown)


src/eos-metrics-instrumentation.c
```diff
@@ -223,6 +235,32 @@ record_os_version (gpointer unused)
...
 
+static gboolean
+is_live_boot (void)
+{
+  g_autofree gchar *cmdline = NULL;
+  g_autoptr(GError) error = NULL;
+
+  if (!g_file_get_contents (KERNEL_CMDLINE_PATH, &cmdline, NULL, &error))
+    {
+      g_warning ("unable to read " KERNEL_CMDLINE_PATH ": %s", error->message);
```
cosimoc: Nitpick: capitalize message

PR #64, opened 2017-01-31
T15063: send eos-image-version xattr as an event

ramcq:
This adds event 6b1c1cfc-bc36-438c-0647-dacd5878f2b3 which is sent once at startup to report the image ID, if present. This is a string such as "eos-eos3.1-amd64-amd64.170115-071322.base" which is saved in an attribute on the root filesystem by the image builder, and allows us to tell the channel that the OS was installed by (eg download, OEM pre-install, Endless hardware, USB stick, etc) and which version was installed. The payload is a single string containing this image ID. Allows machines with deployment-specific images to be found and grouped for metrics analysis purposes.

https://phabricator.endlessm.com/T15063

### DIFF
--- a/data/Makefile.am.inc
+++ b/data/Makefile.am.inc
@@ -36,7 +36,7 @@ edit = sed \
 	-e 's|@instrumentationcachedir[@]|$(instrumentationcachedir)|g' \
 	$(NULL)
 
-tmpfilesddir = $(libdir)/tmpfiles.d/
+tmpfilesddir = $(prefix)/lib/tmpfiles.d/
 tmpfilesd_DATA = data/eos-metrics-instrumentation.conf
 
 EXTRA_DIST += \
--- a/src/eos-metrics-instrumentation.c
+++ b/src/eos-metrics-instrumentation.c
@@ -22,6 +22,7 @@
 #include <glib-object.h>
 #include <glib-unix.h>
 #include <string.h>
+#include <sys/xattr.h>
 
 #include <eosmetrics/eosmetrics.h>
 
@@ -110,6 +111,21 @@
 #define KERNEL_CMDLINE_PATH "/proc/cmdline"
 #define LIVE_BOOT_FLAG_REGEX "\\bendless\\.live_boot\\b"
 #define DUAL_BOOT_FLAG_REGEX "\\bendless\\.image\\.device\\b"
+
+/*
+ * Recorded once at startup to report the image ID. This is a string such as
+ * "eos-eos3.1-amd64-amd64.170115-071322.base" which is saved in an attribute
+ * on the root filesystem by the image builder, and allows us to tell the
+ * channel that the OS was installed by (eg download, OEM pre-install, Endless
+ * hardware, USB stick, etc) and which version was installed. The payload
+ * is a single string containing this image ID, if present.
+ */
+
+#define EOS_IMAGE_VERSION_EVENT "6b1c1cfc-bc36-438c-0647-dacd5878f2b3"
+
+#define EOS_IMAGE_VERSION_XATTR "user.eos-image-version"
+#define EOS_IMAGE_VERSION_PATH "/sysroot"
+#define EOS_IMAGE_VERSION_ALT_PATH "/"
 
 static gboolean prev_time_set = FALSE;
 static gint64 prev_time;
@@ -280,6 +296,70 @@ record_live_boot (gpointer unused)
     else if (is_dual_boot)
       emtr_event_recorder_record_event (emtr_event_recorder_get_default (),
                                         DUAL_BOOT_EVENT, NULL);
+
+    return G_SOURCE_REMOVE;
+}
+
+static gchar *
+get_image_version_for_path (const gchar *path)
+{
+    ssize_t xattr_size;
+    g_autofree gchar *image_version = NULL;
+
+    xattr_size = getxattr (path, EOS_IMAGE_VERSION_XATTR, NULL, 0);
+
+    if (xattr_size < 0 || xattr_size > SSIZE_MAX - 1)
+      return NULL;
+
+    image_version = g_malloc0 (xattr_size + 1);
+
+    xattr_size = getxattr (path, EOS_IMAGE_VERSION_XATTR,
+                           image_version, xattr_size);
+
+    /* this check is primarily for ERANGE, in case the attribute size has
+     * changed from the first call to this one */
+    if (xattr_size < 0)
+      {
+        g_warning ("Error when getting 'eos-image-version' from %s: %s", path,
+                   strerror (errno));
+        return NULL;
+      }
+
+    /* shouldn't happen, but if the filesystem is modified or corrupted, we
+     * don't want to cause assertion errors / D-Bus disconnects with invalid
+     * UTF-8 strings */
+    if (!g_utf8_validate (image_version, xattr_size, NULL))
+      {
+        g_warning ("Invalid UTF-8 when getting 'eos-image-version' from %s",
+                   path);
+        return NULL;
+      }
+
+    return g_steal_pointer (&image_version);
+}
+
+static gchar *
+get_image_version (void)
+{
+    gchar *image_version = get_image_version_for_path (EOS_IMAGE_VERSION_PATH);
+
+    if (image_version == NULL)
+      image_version = get_image_version_for_path (EOS_IMAGE_VERSION_ALT_PATH);
+
+    return image_version;
+}
+
+static gboolean
+record_image_version (gpointer unused)
+{
+    g_autofree gchar *image_version = NULL;
+
+    image_version = get_image_version ();
+
+    if (image_version != NULL)
+      emtr_event_recorder_record_event (emtr_event_recorder_get_default (),
+                                        EOS_IMAGE_VERSION_EVENT,
+                                        g_variant_new_string (image_version));
 
     return G_SOURCE_REMOVE;
 }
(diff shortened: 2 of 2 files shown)


src/eos-metrics-instrumentation.c
```diff
@@ -284,6 +300,70 @@ record_live_boot (gpointer unused)
...
+    image_version = g_malloc0 (xattr_size + 1);
+
+    xattr_size = getxattr (path, EOS_IMAGE_VERSION_XATTR,
+	                       image_version, xattr_size);
+
+    /* this check is just in case the xattr has changed in between the
+     * size checks */
+    if (xattr_size == -1)
+      {
+        g_warning ("Error when getting 'eos-image-version' from %s", path);
```
pwithnall: This warning would be more useful if it appended `strerror (errno)`.

src/eos-metrics-instrumentation.c
```diff
@@ -284,6 +300,70 @@ record_live_boot (gpointer unused)
...
+        return NULL;
```
pwithnall: `image_version` is leaked on this path. You could change the declaration to `g_autofree gchar *image_version = NULL` to simplify the memory handling.

src/eos-metrics-instrumentation.c
```diff
@@ -284,6 +300,70 @@ record_live_boot (gpointer unused)
...
 
+static gchar *
+get_image_version_for_path (const gchar *path)
+{
+    ssize_t xattr_size = 0;
+    gchar *image_version = NULL;
+
+    xattr_size = getxattr (path, EOS_IMAGE_VERSION_XATTR, NULL, 0);
+
+    if (xattr_size == -1)
```
pwithnall: Paranoia mode: `if (xattr_size < 0 || xattr_size > SSIZE_MAX - 1)` instead?

src/eos-metrics-instrumentation.c
```diff
@@ -284,6 +300,70 @@ record_live_boot (gpointer unused)
...
+      return NULL;
+
+    image_version = g_malloc0 (xattr_size + 1);
+
+    xattr_size = getxattr (path, EOS_IMAGE_VERSION_XATTR,
+	                       image_version, xattr_size);
```
pwithnall: Nitpick: Indentation looks a little off on this wrapped line.

src/eos-metrics-instrumentation.c
```diff
@@ -284,6 +300,70 @@ record_live_boot (gpointer unused)
     return G_SOURCE_REMOVE;
 }
 
+static gchar *
+get_image_version_for_path (const gchar *path)
+{
+    ssize_t xattr_size = 0;
```
pwithnall: Nitpick: Misleading to assign a value to this at the moment. If you feel you should, I suggest it’s `-1` to indicate an invalid xattr, rather than `0` to indicate a present-but-zero-length one.

src/eos-metrics-instrumentation.c
```diff
@@ -284,6 +300,70 @@ record_live_boot (gpointer unused)
...
+
+static gchar *
+get_image_version (void)
+{
+    gchar *image_version = get_image_version_for_path (EOS_IMAGE_VERSION_PATH);
+
+    if (image_version == NULL)
+      image_version = get_image_version_for_path (EOS_IMAGE_VERSION_ALT_PATH);
+
+	return image_version;
```
pwithnall: Nitpick: Funky indentation.

data/Makefile.am.inc
```diff
@@ -36,7 +36,7 @@ edit = sed \
 	-e 's|@instrumentationcachedir[@]|$(instrumentationcachedir)|g' \
 	$(NULL)
 
-tmpfilesddir = $(libdir)/tmpfiles.d/
+tmpfilesddir = $(prefix)/lib/tmpfiles.d/
```
ramcq: I don't see how this will break provided systemd's prefix is the same prefix which is provided by the packaging - I was just looking to implement the "common" solution that showed up under https://github.com/search?l=Makefile&q=org%3Aendlessm+tmpfiles&type=Code - for example, systemd itself defines the dir as $(prefix)/lib/tmpfiles.d/ as well.

src/eos-metrics-instrumentation.c
```diff
@@ -111,6 +112,21 @@
 #define LIVE_BOOT_FLAG_REGEX "\\bendless\\.live_boot\\b"
 #define DUAL_BOOT_FLAG_REGEX "\\bendless\\.image\\.device\\b"
 
+/*
+ * Recorded once at startup to report the image ID. This is a string such as
+ * "eos-eos3.1-amd64-amd64.170115-071322.base" which is saved in an attribute
+ * on the root filesystem by the image builder, and allows us to tell the
+ * channel that the OS was installed by (eg download, OEM pre-install, Endless
+ * hardware, USB stick, etc) and which version was installed. The payload
+ * is a single string containing this image ID, if present.
```
ramcq: Seems implicit by the fact that all metrics events must be valid GVariants, so any "string" is defined to be UTF-8.

src/eos-metrics-instrumentation.c
ramcq: Fixed.

data/Makefile.am.inc
pwithnall: OK.

src/eos-metrics-instrumentation.c
ramcq: let's go for xattr_size < 0, I don't see that second-guessing the ssize_t definition is that valuable

src/eos-metrics-instrumentation.c
ramcq: Fixed, typo switching gnome-software tabs to spaces.

src/eos-metrics-instrumentation.c
ramcq: Fixed.

src/eos-metrics-instrumentation.c
ramcq: I used g_autofree originally but it was freeing the return value when the function returned, as far as I could tell...

src/eos-metrics-instrumentation.c
pwithnall: The second part is not second-guessing the `ssize_t` definition — it’s ensuring that the `xattr_size + 1` below doesn’t overflow.

src/eos-metrics-instrumentation.c
pwithnall: `return g_steal_pointer (&image_version);`

src/eos-metrics-instrumentation.c
ramcq: Fixed.

ramcq:
Ready for review again.

src/eos-metrics-instrumentation.c
```diff
@@ -284,6 +300,72 @@ record_live_boot (gpointer unused)
...
+    xattr_size = getxattr (path, EOS_IMAGE_VERSION_XATTR,
+                           image_version, xattr_size);
+
+    /* this check is just in case the xattr has changed in between the
+     * size checks */
+    if (xattr_size < 0)
+      {
+        g_warning ("Error when getting 'eos-image-version' from %s: %s", path,
+                   strerror (errno));
+        g_free (image_version);
```
pwithnall: I think it would be clearer to use `g_autofree gchar *image_version = NULL` and `return g_steal_pointer (&image_version);` on success; but this works.

src/eos-metrics-instrumentation.c
```diff
@@ -284,6 +300,72 @@ record_live_boot (gpointer unused)
...
 
+static gchar *
+get_image_version_for_path (const gchar *path)
+{
+    ssize_t xattr_size;
+    gchar *image_version = NULL;
+
+    xattr_size = getxattr (path, EOS_IMAGE_VERSION_XATTR, NULL, 0);
+
+    if (xattr_size < 0)
```
pwithnall: I think you might as well put the overflow check in. If we’re going to be paranoid enough to check for valid UTF-8, we might as well also check for overflow.

src/eos-metrics-instrumentation.c
ramcq: Oops. Additional coffee required. Fixed!

src/eos-metrics-instrumentation.c
```diff
@@ -284,6 +300,72 @@ record_live_boot (gpointer unused)
...
+      return NULL;
+
+    image_version = g_malloc0 (xattr_size + 1);
+
+    xattr_size = getxattr (path, EOS_IMAGE_VERSION_XATTR,
+                           image_version, xattr_size);
+
+    /* this check is just in case the xattr has changed in between the
+     * size checks */
```
pwithnall: Hmm, thinking about this again, we should be checking that `xattr_size` from the second call is equal to (or less than) `xattr_size` from the first call; as well as that it’s positive. Otherwise the xattr value could increase in size between the two calls, and we would end up with a truncated copy of it in `image_version`, and an `xattr_size` which is now longer than the allocated area for `image_version`.

src/eos-metrics-instrumentation.c
ramcq: Ooh, TIL. Fixed. :)

src/eos-metrics-instrumentation.c
ramcq: Fixed by clarifying the vague comment - in the case the buffer is truncated, -1 will be returned with ERANGE set.

pwithnall: I’m wrong: `getxattr()` would return `-1` and `ERANGE` in that case. Ship it.

PR #66, opened 2017-05-05
T16934: add new network ID event to group machines by location

ramcq:
Introduce the event `38eb48f8-e131-9b57-77c6-35e0590c82fd` which is recorded at system startup and when we detect a change in the default route.

The auxiliary payload is a 32-bit unsigned integer containing a hash of the ethernet MAC address of the gateway, favouring IPv4 if available, or IPv6 if not. The intention is to provide a value which is opaque and stable which is the same for every system located on the same physical network.

The new `eos-network-id` command prints this ID for user inspection.

https://phabricator.endlessm.com/T16934

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 # Build products
 /eos-metrics-instrumentation
+/eos-network-id
 /tests/test-persistent-tally
 /data/eos-metrics-instrumentation.conf
 /geoclue.[ch]
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,7 @@ stamp*
 .dirstamp
 *.tar.*
 *.pc
+Makecache
 Makefile
 Makefile.in
 gtk-doc.make
--- a/src/eins-network-id.c
+++ b/src/eins-network-id.c
@@ -1,0 +1,267 @@
+/* Copyright 2017 Endless Mobile, Inc.
+ *
+ * This file is part of eos-metrics-instrumentation.
+ *
+ * eos-metrics-instrumentation is free software: you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 2 of the License, or (at your
+ * option) any later version.
+ *
+ * eos-metrics-instrumentation is distributed in the hope that it will be
+ * useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
+ * Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with eos-metrics-instrumentation.  If not, see
+ * <http://www.gnu.org/licenses/>.
+ */
+
+#include <arpa/inet.h>
+#include <errno.h>
+#include <glib.h>
+#include <stdio.h>
+
+#include "eins-network-id.h"
+
+#define PROC_IPV4_ROUTE_FILE "/proc/net/route"
+#define IPV4_ROUTE_REGEXP "^(?:\\S+\\s+){2}(?P<Gateway>[0-9A-Fa-f]{8})\\s+(?:\\S+\\s+){4}(?P<Mask>00000000)"
+
+/*
+ * greps the /proc/net/route file for the IPv4 default route. note that even
+ * on machines with multiple interfaces, the first listed 00000000 route will
+ * be the first one selected for outbound traffic, so we don’t need to do any
+ * more complex searching/sorting here (verified in net/ipv4/fib_trie.c on
+ * kernel 4.10.5 - the trie is ordered by (prefix,tos,priority) so the highest
+ * preference default route will appear first)
+ */
+static gchar *
+get_ipv4_default_gateway (void)
+{
+  g_autofree gchar *route = NULL;
+  g_autoptr(GError) err = NULL;
+
+  if (!g_file_get_contents (PROC_IPV4_ROUTE_FILE, &route, NULL, &err))
+    {
+      g_debug ("could not load " PROC_IPV4_ROUTE_FILE ": %s", err->message);
+      return NULL;
+    }
+
+  g_autoptr(GRegex) route_rx = g_regex_new (IPV4_ROUTE_REGEXP, G_REGEX_MULTILINE, 0, NULL);
+  g_assert (route_rx != NULL);
+
+  g_autoptr(GMatchInfo) route_mi = NULL;
+  if (!g_regex_match (route_rx, route, 0, &route_mi))
+    {
+      g_debug ("couldn’t match IPv4 default gateway in " PROC_IPV4_ROUTE_FILE);
+      return NULL;
+    }
+
+  const gchar *gateway = g_match_info_fetch_named (route_mi, "Gateway");
+  g_assert (gateway != NULL);
+
+  /* the input to this g_ascii_strtoull call should already be
+   * rigorously validated by the regexp: failure is not an option */
+  gchar *endptr = NULL;
+  errno = 0;
+  guint32 gateway_ip = (guint32) g_ascii_strtoull (gateway, &endptr, 16);
+  g_assert (errno == 0);
+  g_assert (gateway + 8 == endptr);
+  gateway_ip = htonl (gateway_ip);
+
+  return g_strdup_printf ("%u.%u.%u.%u",
+                          gateway_ip >> 24,
+                          0xFF & (gateway_ip >> 16),
+                          0xFF & (gateway_ip >> 8),
+                          0xFF & gateway_ip);
+}
+
+#define PROC_IPV4_ARP_FILE "/proc/net/arp"
+#define IPV4_ARP_REGEXP "^%s\\s+(\\S+\\s+){2}(?P<HW>[0-9A-Fa-f:]{17})"
+
+/*
+ * find the given IPv4 address in the ARP cache to find the HW address.
+ * relies on two assumptions:
+ *  - the first (obviously) is that the IP has been communicated with recently
+ *    enough to be in the cache. for an internet-connected host, this is a
+ *    pretty safe bet for the gateway IP.
+ *  - the second is that if the host has multiple interfaces which are in the
+ *    same IPv4 segment, that these are the same physical network, because the
+ *    search is not constrained by interface which would be the strictly
+ *    correct thing to do. this configuration seems pretty unlikely (two bonded
+ *    load-balanced links with separate gateways?) on a client machine, and for
+ *    our purposes it doesn’t matter if we always report one gateway, or report
+ *    both over time, so it’s not worth getting excited about.
+ */
+static gchar *
+get_ipv4_hwaddr (const gchar *address)
+{
+  g_autofree gchar *arp = NULL;
+  g_autoptr(GError) err = NULL;
+
+  if (!g_file_get_contents (PROC_IPV4_ARP_FILE, &arp, NULL, &err))
+    {
+      g_debug ("could not load " PROC_IPV4_ARP_FILE ": %s", err->message);
+      return NULL;
+    }
+
+  g_autofree gchar *address_esc = g_regex_escape_string (address, -1);
+  g_autofree gchar *arp_rx_raw = g_strdup_printf (IPV4_ARP_REGEXP, address_esc);
+  g_autoptr(GRegex) arp_rx = g_regex_new (arp_rx_raw, G_REGEX_MULTILINE, 0, NULL);
+  g_assert (arp_rx != NULL);
+
+  g_autoptr(GMatchInfo) arp_mi = NULL;
+  if (!g_regex_match (arp_rx, arp, 0, &arp_mi))
+    {
+      g_debug ("couldn’t find HW address in " PROC_IPV4_ARP_FILE);
+      return NULL;
+    }
+
+  const gchar *hw = g_match_info_fetch_named (arp_mi, "HW");
+  g_assert (hw != NULL);
+  return g_ascii_strdown (hw, -1);
+}
+
+/*
+ * the format of each line is:
+ * <dest 32> <prefix 2> <src 32> <prefix 2> <gateway 32> \
+ *  <metric 8> <refcount 8> <use? 8> <flags 8> <if>
+ * we want to match dest/prefix of 0/0, any src/prefix, capture the gateway,
+ * any metric/ref/use, and match flags ending 3, meaning RTF_UP + RTF_GATEWAY
+ */
+#define PROC_IPV6_ROUTE_FILE "/proc/net/ipv6_route"
+#define IPV6_ROUTE_REGEXP "^0{32}\\s00\\s[0-9A-Fa-f]{32}\\s[0-9A-Fa-f]{2}\\s(?P<Gateway>[0-9A-Fa-f]{32})\\s([0-9A-Fa-f]{8}\\s){3}[0-9A-Fa-f]{7}3"
+
+/*
+ * greps the /proc/net/ipv6_route file for the first IPv6 default route. same
+ * assumptions as above apply.
+ */
+static gchar *
+get_ipv6_default_gateway (void)
+{
+  g_autofree gchar *route = NULL;
+  g_autoptr(GError) err = NULL;
+
+  if (!g_file_get_contents (PROC_IPV6_ROUTE_FILE, &route, NULL, &err))
+    {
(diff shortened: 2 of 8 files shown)


pwithnall:
Some drive-by thoughts before I actually look at the code:
 * Given that IPv6 can represent IPv4 and IPv6 addresses, maybe it would make more sense to construct and hash an IPv6 address? If an IPv6 address is available for the gateway, use that. If only an IPv4 address is available, use the mapped IPv6 address.
 * What about multi-homed machines?

src/eins-network-id.c
```diff
@@ -0,0 +1,238 @@
...
+ * preference default route will appear first)
+ */
+gchar *
+get_ipv4_default_gateway (void)
+{
+  g_autofree gchar *route = NULL;
+
+  if (!g_file_get_contents (PROC_IPV4_ROUTE_FILE, &route, NULL, NULL) || !route)
+    {
+      g_debug ("could not load " PROC_IPV4_ROUTE_FILE);
```
pwithnall: For debugging, it would be nice to use a `GError` and print the `error->message` in this debug message.

src/eins-network-id.c
```diff
@@ -0,0 +1,238 @@
...
+      return NULL;
+    }
+
+  g_autoptr(GRegex) route_rx = g_regex_new (IPV4_ROUTE_REGEXP, G_REGEX_MULTILINE, 0, NULL);
```
pwithnall: This doesn’t handle errors. If you’re confident no error could ever occur, it would still be good to have a `g_assert (route_rx != NULL)` just below.

src/eins-network-id.c
```diff
@@ -0,0 +1,238 @@
...
+
+  if (!g_regex_match (route_rx, route, 0, &route_mi) || !route_mi)
+    {
+      g_debug ("couldn't match IPv4 default gateway in " PROC_IPV4_ROUTE_FILE);
+      return NULL;
+    }
+
+  const gchar *gateway = g_match_info_fetch_named (route_mi, "Gateway");
+  g_assert (gateway);
+  guint64 gateway_ip = g_ascii_strtoull (gateway, NULL, 16);
```
pwithnall: This needs some error handling.

src/eins-network-id.c
```diff
@@ -0,0 +1,238 @@
...
+ *    both over time, so it's not worth getting excited about.
+ */
+gchar *
+get_ipv4_hwaddr (const gchar *address)
+{
+  g_autofree gchar *arp = NULL;
+
+  if (!g_file_get_contents (PROC_IPV4_ARP_FILE, &arp, NULL, NULL) || !arp)
+    {
+      g_debug ("could not load " PROC_IPV4_ARP_FILE);
```
pwithnall: Same comments as above apply in this function too.

ramcq:
We are not hashing IP addresses, we are hashing the MAC address of the gateway/router that provides the default route. The IPv4 routes don't appear in the IPv6 routing table, so we have to examine both stacks separately to determine the default route, and then look up the gateway in the ARP table (for IPv4) or neighbour discovery cache (for IPv6).

Multi-homed machines are discussed in comments in the code, essentially - unless you are load-balancing across bonded routes, which I consider very unlikely for a client machine, interface weights (eg wireless is more expensive / less preferable than wired) mean you will in fact have a stable routing table and for each program, the default route will only go to one place.

src/eins-network-id.c
```diff
@@ -0,0 +1,238 @@
...
+      return NULL;
+    }
+
+  g_autofree gchar *arp_rx_raw = g_strdup_printf (IPV4_ARP_REGEXP, address);
```
pwithnall: Technically you should call `g_regex_escape_string()` on `address` first.

src/eins-network-id.c
```diff
@@ -0,0 +1,238 @@
...
+  g_autoptr(GRegex) arp_rx = g_regex_new (arp_rx_raw, G_REGEX_MULTILINE, 0, NULL);
```
pwithnall: Needs a `g_assert (arp_rx != NULL)` as in the other function.

src/eins-network-id.c
```diff
@@ -0,0 +1,238 @@
...
+ * <http://www.gnu.org/licenses/>.
+ */
+
+#include <glib.h>
+#include <stdio.h>
+
+#include "eins-network-id.h"
+
+#define PROC_IPV4_ROUTE_FILE "/proc/net/route"
+#define IPV4_ROUTE_REGEXP "^(?:\\S+\\s+){2}(?P<Gateway>[0-9A-F]{8})\\s+(?:\\S+\\s+){4}(?P<Mask>00000000)"
```
pwithnall: Do you also want to allow `a-f` (lowercase) for the `<Gateway>`? This might not be the format which appears in the kernel at the moment, but it seems wise to be lenient with the parsing here.

src/eins-network-id.c
```diff
@@ -0,0 +1,238 @@
...
+      return NULL;
+    }
+
+  const gchar *hw = g_match_info_fetch_named (arp_mi, "HW");
+  g_assert (hw);
+  return g_ascii_strdown (hw, -1);
+}
+
+#define PROC_IPV6_ROUTE_FILE "/proc/net/ipv6_route"
+#define IPV6_ROUTE_REGEXP "^(0{32}\\s00\\s){2}(?P<Gateway>[0-9a-f]{32})"
```
pwithnall: Similarly, would it make sense to allow `A-F` for the `<Gateway>`?

src/eins-network-id.c
```diff
@@ -0,0 +1,238 @@
...
+
+  const gchar *gateway = g_match_info_fetch_named (route_mi, "Gateway");
+  g_assert (gateway);
+
+  guint32 addr[8];
+  if (sscanf(gateway,"%4x%4x%4x%4x%4x%4x%4x%4x",
+             &addr[0], &addr[1], &addr[2], &addr[3],
+             &addr[4], &addr[5], &addr[6], &addr[7]) < 6)
+    {
+      g_assert_not_reached ();
```
pwithnall: This could do with a comment saying that the regexp has already validated the format of `gateway`, so this should always succeed.

src/eins-network-id.c
```diff
@@ -0,0 +1,238 @@
...
+  const gchar *hw = g_match_info_fetch_named (arp_mi, "HW");
+  g_assert (hw);
+  return g_ascii_strdown (hw, -1);
+}
+
+#define PROC_IPV6_ROUTE_FILE "/proc/net/ipv6_route"
+#define IPV6_ROUTE_REGEXP "^(0{32}\\s00\\s){2}(?P<Gateway>[0-9a-f]{32})"
+
+/*
+ * greps the /proc/net/ipv6_route file for the first IPv6 default route. same
```
pwithnall: Does it make sense to go for the *first* default route we find, or do something a bit more reproducible, like sorting by address? Arbitrary, but reproducible. I suspect the order of the `ipv6_route` file might depend on the order in which interfaces are brought up, which is essentially random. I haven’t checked any of this though — just asking questions.

src/eins-network-id.c
```diff
@@ -0,0 +1,238 @@
...
+
+/*
+ * no IPv6 equivalent of /proc/net/arp exists to view the neigh cache, so rather
+ * than breaking out netlink, we invoke the "ip" command for sanity's sake.
+ */
+gchar *
+get_ipv6_hwaddr (const gchar *address)
+{
+  gchar *neigh_cmd[] = { "ip", "-6", "neigh", "show", (gchar *) address, NULL };
+  g_autofree gchar *neigh = NULL;
```
pwithnall: 🏇 ??!

src/eins-network-id.c
```diff
@@ -0,0 +1,238 @@
...
+#define IPV6_NEIGH_REGEXP "^[0-9a-f:]+\\s(\\S+\\s+){2}lladdr\\s(?P<HW>[0-9a-f:]{17})"
+
+/*
+ * no IPv6 equivalent of /proc/net/arp exists to view the neigh cache, so rather
+ * than breaking out netlink, we invoke the "ip" command for sanity's sake.
+ */
+gchar *
+get_ipv6_hwaddr (const gchar *address)
+{
+  gchar *neigh_cmd[] = { "ip", "-6", "neigh", "show", (gchar *) address, NULL };
```
pwithnall: This could be `const gchar *[]`.

src/eins-network-id.h
```diff
@@ -0,0 +1,27 @@
...
+ * along with eos-metrics-instrumentation.  If not, see
+ * <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef EINS_NETWORK_ID_H
+#define EINS_NETWORK_ID_H
+
+#include <glib.h>
+
+gboolean eins_network_id_get (guint32 *);
```
pwithnall: Various bits of GLib tooling generally prefer it if you put the argument name in function declarations.

src/eins-network-id.c
```diff
@@ -0,0 +1,238 @@
...
+ * along with eos-metrics-instrumentation.  If not, see
+ * <http://www.gnu.org/licenses/>.
+ */
+
+#include <glib.h>
+#include <stdio.h>
+
+#include "eins-network-id.h"
+
+#define PROC_IPV4_ROUTE_FILE "/proc/net/route"
```
pwithnall: Do we need to worry about the endianness of any of these files? If two machines with different endianness are on the same network, we presumably want them to report the same network ID.

src/eins-network-id.c
```diff
@@ -0,0 +1,238 @@
...
+#define IPV4_ROUTE_REGEXP "^(?:\\S+\\s+){2}(?P<Gateway>[0-9A-F]{8})\\s+(?:\\S+\\s+){4}(?P<Mask>00000000)"
+
+/*
+ * greps the /proc/net/route file for the IPv4 default route. note that even
+ * on machines with multiple interfaces, the first listed 00000000 route will
+ * be the first one selected for outbound traffic, so we don't need to do any
```
pwithnall: s/'/’/

:-)

src/eins-network-id.c
ramcq: That's a good Q, although given the number of architectures in the universe is trending down with the exception of backwards-compatibility with two little-endian mainstays, may not be all that critical for Endless. For the sake of completeness I'll check it out. This file definitely has the IPs as little-endian, but I don't know if that's because it's host-endian or because the trie has them backwards for some implementation reason. I don't want to stick it through htonl only to find it comes out in the same order on big-endian machines too! :D

src/eins-network-id.c
ramcq: Sorted.

src/eins-network-id.c
ramcq: It's indeed host-endian - htonl added. :)

src/eins-network-id.c
ramcq: Sorted.

src/eins-network-id.c
ramcq: Sorted.

src/eins-network-id.c
ramcq: Sorted.

src/eins-network-id.c
ramcq: Sorted.

src/eins-network-id.c
ramcq: Sorted.

src/eins-network-id.c
ramcq: Sorted.

src/eins-network-id.c
ramcq: Sorted.

src/eins-network-id.c
ramcq: I checked, as per the IPv4 route file it's a walk of the tree which is ordered by of prefix and metric which should be stable enough - ethernet always beats wifi, for example. The first "hit" that the kernel makes when routing a packet via the gateway will be our first hit too.

Although, in checking this I did find a bug in picking up routes for 0000::/0 which were not via gateways, so I extended the regexp to check the flags show the route is RTF_UP (0x1) and involves a RTF_GATEWAY (0x2). RTF_HOST (0x4) should be mutually exclusive with the default youre, and RTF_REINSTATE (0x8) is unused in Linux so I don't think we need to worry about being more clever than looking for a trailing 3.

src/eins-network-id.c
ramcq: 👍

src/eins-network-id.c
ramcq: 👍

src/eins-network-id.c
ramcq: 🏇 replaced with ndisc. :P

src/eins-network-id.h
ramcq: :+1:

src/eins-network-id.c
ramcq: 💯

src/eins-network-id.c
```diff
@@ -0,0 +1,260 @@
...
+      return NULL;
+    }
+
+  g_autoptr(GRegex) route_rx = g_regex_new (IPV4_ROUTE_REGEXP, G_REGEX_MULTILINE, 0, NULL);
+  g_assert (route_rx != NULL);
+
+  g_autoptr(GMatchInfo) route_mi = NULL;
+  if (!g_regex_match (route_rx, route, 0, &route_mi))
+    {
+      g_debug ("couldn’t match IPv4 default gateway in " PROC_IPV4_ROUTE_FILE);
```
pwithnall: Secret additional curly apostrophe! :+1: 🎉

src/eins-network-id.c
```diff
@@ -0,0 +1,260 @@
...
+  const gchar *gateway = g_match_info_fetch_named (route_mi, "Gateway");
+  g_assert (gateway != NULL);
+
+  /* the input to this g_ascii_strtoull call should already be
+   * rigorously validated by the regexp: failure is not an option */
+  gchar *endptr = NULL;
+  errno = 0;
+  guint32 gateway_ip = (guint32) g_ascii_strtoull (gateway, &endptr, 16);
+  g_assert (gateway + 8 == endptr);
+  g_assert (gateway_ip != 0 || errno == 0);
```
pwithnall: Technically the check for `errno` should be before the other `g_assert()`, just in case `g_assert()` resets the `errno` (it shouldn’t, but `errno` is fickle).

pwithnall: Also, I think you should drop the `gateway_ip != 0` clause: you should only care about `errno` and `endptr` for the error reporting, since the return value can be `0` *or* `G_MAXUINT64` in different error cases.

i.e. `g_ascii_strtoull()` is a terribly designed function.

src/eins-network-id.c
```diff
@@ -0,0 +1,260 @@
...
+  g_autofree gchar *ndisc = NULL;
+  g_autofree gchar *ndisc_err = NULL;
+  g_autoptr(GError) err = NULL;
+  gint ret = 0;
+
+  if (!g_spawn_sync (NULL, (gchar **) &ndisc_cmd[0], NULL,
+                     G_SPAWN_SEARCH_PATH, NULL, NULL,
+                     &ndisc, &ndisc_err, &ret, &err) || ret)
+    {
+      g_debug ("could not execute ip -6 neigh command (ret %i): %s", ret,
```
pwithnall: Would be good to format `ndisc_cmd` and print it in this debug message properly, so the `address` is in the output and it’s completely reproducible from a terminal.

src/eins-network-id.c
```diff
@@ -0,0 +1,260 @@
...
+      return NULL;
+    }
+
+  g_autoptr(GRegex) ndisc_rx = g_regex_new (IPV6_NDISC_REGEXP, G_REGEX_MULTILINE, 0, NULL);
+  g_assert (ndisc_rx != NULL);
+
+  g_autoptr(GMatchInfo) ndisc_mi = NULL;
+  if (!g_regex_match (ndisc_rx, ndisc, 0, &ndisc_mi))
+    {
+      g_debug ("couldn’t find HW address in ip -6 ndisc output: %s", ndisc);
```
pwithnall: Similarly, it would be good to format `ndisc_cmd` into this debug message.

.gitignore
```diff
@@ -23,6 +23,7 @@ stamp*
 .dirstamp
 *.tar.*
 *.pc
+Makecache
```
pwithnall: I suggest using [`git.mk`](https://github.com/behdad/git.mk), but that can happen separately as a future PR.

PR #69, opened 2017-07-21
Report whether Windows licenses are baked into the firmware

wjt:
https://phabricator.endlessm.com/T18296

### DIFF
--- a/src/eos-metrics-instrumentation.c
+++ b/src/eos-metrics-instrumentation.c
@@ -146,6 +146,31 @@
 #define EOS_IMAGE_VERSION_PATH "/sysroot"
 #define EOS_IMAGE_VERSION_ALT_PATH "/"
 
+/*
+ * Reported once at startup to describe whether certain ACPI tables are present
+ * on the system. The payload has type a{sb} -- a map from string to boolean.
+ * The tables in question are MSDM and SLIC, which hold OEM Windows license
+ * information on newer and older systems respectively. Example payloads:
+ *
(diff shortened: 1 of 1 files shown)


src/eos-metrics-instrumentation.c
```diff
@@ -946,6 +971,33 @@ network_dbus_proxy_new (void)
 }
 
 static gboolean
+record_windows_licenses (gpointer unused)
+{
+    g_autoptr(GFile) tables = g_file_new_for_path (ACPI_TABLES_PATH);
+    g_auto(GVariantBuilder) builder;
+    gsize i;
+
+    g_variant_builder_init (&builder, G_VARIANT_TYPE ("a{sb}"));
```
ptomato: Drive-by comment — consider using `bb` as the variant type so that we aren't storing hundreds of thousands of copies of the strings `SLIC` and `MSDM` in the database

rahmanShettia: Although we would have  many of these strings a dictionary like {"MSDM": true, "SLIC": false} would be much feasible to data module as the rest of the dictionary like events

ramcq: After discussions with @rahmanShettia we decided to go ahead with the self-documenting format on the basis that the metrics server will need to gain the ability to compress or normalise events as it scales in any case.

ptomato: It also wastes user bandwidth, but I think we can get away with O(10 bytes) per user per boot

PR #70, opened 2017-07-22
change Windows license event to a bitfield

ramcq:
Alright... bikeshed challenge accepted...

Given the most interesting values here are 0 or non-zero, this
lends itself to a bitfield quite well.
 0: No OEM Windows
 1: MSDM (Windows >= 8)
 2: SLIC (Windows Vista-ish)
 3: Unexpected but could happen

Change the event ID due to the new signature.

https://phabricator.endlessm.com/T18296

### DIFF
--- a/src/eos-metrics-instrumentation.c
+++ b/src/eos-metrics-instrumentation.c
@@ -148,28 +148,29 @@
 
 /*
  * Reported once at startup to describe whether certain ACPI tables are present
- * on the system. The payload has type a{sb} -- a map from string to boolean.
- * The tables in question are MSDM and SLIC, which hold OEM Windows license
- * information on newer and older systems respectively. Example payloads:
+ * on the system. The payload has type u, formed as a bitmask of which ACPI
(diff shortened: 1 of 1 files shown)


src/eos-metrics-instrumentation.c
```diff
@@ -167,9 +167,9 @@
  *    false   |   false   | Endless OS is the sole OS, PC came without Windows
  *    false   |   true    | Dual-booting with a retail Windows
  */
-#define WINDOWS_LICENSE_TABLES_EVENT "5311a1be-0ed1-44c2-b780-7c4d5d6037f8"
+#define WINDOWS_LICENSE_TABLES_EVENT "ef74310f-7c7e-ca05-0e56-3e495973070a"
```
wjt: Please update the comment above to reflect the new payload type.

ptomato:
Heh :smile:
:+1: modulo Will's comment about the comment. (Also make sure to add it to the wiki page where all the recorded events are described.)

PR #71, opened 2017-09-06
T18444

smspillaz:
https://phabricator.endlessm.com/T18444

### DIFF
--- a/src/Makefile.am.inc
+++ b/src/Makefile.am.inc
@@ -16,7 +16,10 @@
 ## along with eos-metrics-instrumentation.  If not, see
 ## <http://www.gnu.org/licenses/>.
 
-libexec_PROGRAMS = src/eos-metrics-instrumentation
+libexec_PROGRAMS = \
+	src/eos-metrics-instrumentation \
+	src/eos-crash-metrics \
+	$(NULL)
 
 src/geoclue.c: src/geoclue.h
 src/geoclue.h: $(GEOCLUE_DBUS_INTERFACE_XML)
@@ -45,3 +48,14 @@ src_eos_metrics_instrumentation_CPPFLAGS = \
 
 # Libraries to link against
 src_eos_metrics_instrumentation_LDADD = $(EOS_INSTRUMENTATION_LIBS)
+
+src_eos_crash_metrics_SOURCES = \
+	src/eos-crash-metrics.c \
+	$(NULL)
+
+
+src_eos_crash_metrics_CPPFLAGS = \
+	$(EOS_INSTRUMENTATION_CFLAGS) \
+	$(NULL)
+
+src_eos_crash_metrics_LDADD = $(EOS_INSTRUMENTATION_LIBS)
--- a/src/eos-crash-metrics.c
+++ b/src/eos-crash-metrics.c
@@ -1,0 +1,194 @@
+/* Copyright 2017 Endless Mobile, Inc.
+ *
+ * This file is part of eos-metrics-instrumentation.
+ *
+ * eos-metrics-instrumentation is free software: you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 2 of the License, or (at your
+ * option) any later version.
+ *
+ * eos-metrics-instrumentation is distributed in the hope that it will be
+ * useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
+ * Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with eos-metrics-instrumentation.  If not, see
+ * <http://www.gnu.org/licenses/>.
+ */
+
+#include <string.h>
+#include <stdlib.h>
+#include <eosmetrics/eosmetrics.h>
+#include <ostree.h>
+
+#define PROGRAM_DUMPED_CORE_EVENT "ed57b607-4a56-47f1-b1e4-5dc3e74335ec"
+#define EXPECTED_NUMBER_ARGS 3
+
+static void
+report_crash (const char *binary,
+              gint16 signal,
+              gint64 timestamp,
+              const char *ostree_commit,
+              const char *ostree_url)
+{
+  GVariantDict dict;
+  g_variant_dict_init (&dict, NULL);
+
+  g_variant_dict_insert_value (&dict, "binary", g_variant_new_string (binary));
+  g_variant_dict_insert_value (&dict, "signal", g_variant_new_int16 (signal));
+  g_variant_dict_insert_value (&dict, "timestamp", g_variant_new_int16 (timestamp));
+  g_variant_dict_insert_value (&dict, "ostree_commit", g_variant_new_string (ostree_commit));
+  g_variant_dict_insert_value (&dict, "ostree_url", g_variant_new_string (ostree_url));
+
+  emtr_event_recorder_record_event_sync (emtr_event_recorder_get_default (),
+                                         PROGRAM_DUMPED_CORE_EVENT,
+                                         g_variant_dict_end (&dict));
+}
+
+static OstreeSysroot *
+load_ostree_sysroot (GError **error)
+{
+  OstreeSysroot *sysroot = ostree_sysroot_new_default ();
+  if (!ostree_sysroot_load (sysroot, NULL, error))
+    return NULL;
+
+  return sysroot;
+}
+
+static char *
+get_eos_ostree_repo_url (OstreeRepo *repo)
+{
+  GKeyFile *config = NULL;
+  g_autoptr(GError) error = NULL;
+  char *url = NULL;
+
+  config = ostree_repo_get_config (repo);
+
+  if (!(url = g_key_file_get_string (config, "remote \"eos\"", "url", &error)))
+    {
+      g_warning ("Unable to read OSTree config for eos remote URL: %s", error->message);
+      return NULL;
+    }
+
+  return url;
+}
+
+static char *
+get_eos_ostree_deployment_commit (OstreeSysroot *sysroot, OstreeRepo *repo)
+{
+  OstreeDeployment *deployment = ostree_sysroot_get_booted_deployment (sysroot);
+  GKeyFile *config = NULL;
+  g_autoptr(GError) error = NULL;
+  g_autofree char *refspec = NULL;
+  char *commit = NULL;
+
+  if (!deployment)
+    {
+      g_warning ("OSTree deployment is not currently booted, cannot read state");
+      return NULL;
+    }
+
+  config = ostree_deployment_get_origin (deployment);
+
+  if (!(refspec = g_key_file_get_string (config, "origin", "refspec", &error)))
+    {
+      g_warning ("Unable to read OSTree refspec for booted deployment: %s", error->message);
+      return NULL;
+    }
+
+  if (!ostree_repo_resolve_rev (repo, refspec, FALSE, &commit, &error))
+    {
+      g_warning ("Unable to resolve revision for refpsec %s: %s", refspec, error->message);
+      return NULL;
+    }
+
+  return commit;
+}
+
+static const char *blacklisted_prefixes[] =
+{
+  "/home",
+  "/sysroot/home"
+};
+
+static gboolean
+is_blacklisted_path (const char *path)
+{
+  const size_t n_blacklisted_prefixes = G_N_ELEMENTS (blacklisted_prefixes);
+  size_t i = 0;
+
+  for (; i < n_blacklisted_prefixes; ++i)
+    {
+      if (g_str_has_prefix (path, blacklisted_prefixes[i]))
+        return TRUE;
+    }
+
+  return FALSE;
+}
+
+/* The kernel gives us paths in the format !usr!bin!myprogram */
+static void
+normalize_path (char *path)
+{
+  size_t i = 0;
+
+  for (; i < strlen (path); ++i)
+    path[i] = path[i] == '!' ? '/' : path[i];
+}
+
+
+int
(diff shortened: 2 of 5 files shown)


src/eos-crash-metrics.c
```diff
@@ -0,0 +1,183 @@
...
+{
+  "!home",
+  "!sysroot!home"
+};
+
+static gboolean
+is_blacklisted_path (const char *path)
+{
+  /* I can never remember that helper macro that does this ... */
+  const size_t n_blacklisted_prefixes = sizeof (blacklisted_prefixes) / sizeof (blacklisted_prefixes[0]);
```
cosimoc: `G_N_ELEMENTS()`

src/eos-crash-metrics.c
```diff
@@ -0,0 +1,183 @@
...
+  g_autoptr(OstreeSysroot) sysroot = NULL;
+  g_autoptr(OstreeRepo) repo = NULL;
+  g_autoptr(GError) error = NULL;
+  g_autofree char *url = NULL;
+  g_autofree char *commit = NULL;
+
+  if (is_blacklisted_path (path))
+    {
+      g_message ("%s is blacklisted, not reporting crash", path);
+      return 0;
```
cosimoc: `return EXIT_SUCCESS`

src/eos-crash-metrics.c
```diff
@@ -0,0 +1,183 @@
...
+      commit = get_eos_ostree_deployment_commit (sysroot, repo);
+    }
+  else
+    {
+      g_warning ("Unable to read ostree repo from sysroot: %s", error->message);
+    }
+
+  report_crash (path, signal, timestamp, url, commit);
+
+  return 0;
```
cosimoc: See above

src/eos-crash-metrics.c
```diff
@@ -0,0 +1,183 @@
...
+
+  if (!deployment)
+    {
+      g_warning ("OSTree deployment is not currently booted, cannot read state");
+      return NULL;
+    }
+
+  config = ostree_deployment_get_origin (deployment);
+
+  if (!ostree_sysroot_get_repo (sysroot, &repo, NULL, &error))
```
cosimoc: You already have the repo object passed-in as a parameter.

src/eos-crash-metrics.c
```diff
@@ -0,0 +1,183 @@
...
+      return NULL;
+    }
+
+  if (!ostree_repo_resolve_rev (repo, refspec, FALSE, &commit, &error))
+    {
+      g_warning ("Unable to resolve revision for refpsec %s: %s", refspec, error->message);
+      return NULL;
+    }
+
+  return commit;
```
cosimoc: You need to return `g_steal_pointer (&commit)`, otherwise `commit` will be freed when leaving the scope

src/eos-crash-metrics.c
```diff
@@ -0,0 +1,183 @@
...
+    {
+      g_warning ("Unable to get current OSTree sysroot: %s", error->message);
+      return NULL;
+    }
+
+  return sysroot;
+}
+
+static char *
+get_eos_ostree_repo_url (OstreeSysroot *sysroot, OstreeRepo *repo)
```
cosimoc: The sysroot is not used here.

src/eos-crash-metrics.c
```diff
@@ -0,0 +1,183 @@
...
+  g_autoptr(GHashTable) ht = g_hash_table_new_full (g_str_hash,
+                                                    g_str_equal,
+                                                    g_free,
+                                                    (GDestroyNotify) g_variant_unref);
+
+  g_hash_table_insert (ht, g_strdup ("binary"), g_variant_new_string (binary));
+  g_hash_table_insert (ht, g_strdup ("signal"), g_variant_new_int16 (signal));
+  g_hash_table_insert (ht, g_strdup ("timestamp"), g_variant_new_int16 (timestamp));
+  g_hash_table_insert (ht, g_strdup ("ostree_commit"), g_variant_new_string (ostree_commit != NULL ? ostree_commit : ""));
+  g_hash_table_insert (ht, g_strdup ("ostree_commit"), g_variant_new_string (ostree_url != NULL ? ostree_url : ""));
```
cosimoc: This should be `ostree_url` no?

src/eos-crash-metrics.c
```diff
@@ -0,0 +1,183 @@
...
+{
+  g_autoptr(GHashTable) ht = g_hash_table_new_full (g_str_hash,
+                                                    g_str_equal,
+                                                    g_free,
+                                                    (GDestroyNotify) g_variant_unref);
+
+  g_hash_table_insert (ht, g_strdup ("binary"), g_variant_new_string (binary));
+  g_hash_table_insert (ht, g_strdup ("signal"), g_variant_new_int16 (signal));
+  g_hash_table_insert (ht, g_strdup ("timestamp"), g_variant_new_int16 (timestamp));
+  g_hash_table_insert (ht, g_strdup ("ostree_commit"), g_variant_new_string (ostree_commit != NULL ? ostree_commit : ""));
```
cosimoc: In case the ostree commit and URL are not available, I don't think we should report the crash, because the system most likely has been tampered with.

src/eos-crash-metrics.c
```diff
@@ -0,0 +1,183 @@
...
+  g_hash_table_insert (ht, g_strdup ("ostree_commit"), g_variant_new_string (ostree_url != NULL ? ostree_url : ""));
+
+  g_message("%s %s", ostree_commit, ostree_url);
```
cosimoc: I think this could be removed. Otherwise, please make the message more descriptive, and add a space before paren

src/eos-crash-metrics.c
```diff
@@ -0,0 +1,181 @@
+/* Copyright 2017 Endless Mobile, Inc. */
+
+#include <stdlib.h>
+#include <eosmetrics/eosmetrics.h>
+#include <ostree.h>
+
+#define PROGRAM_DUMPED_CORE_EVENT "ed57b607-4a56-47f1-b1e4-5dc3e74335ec"
+
+static GVariant *
+hashtable_to_variant (GHashTable *ht)
```
cosimoc: Instead of this function, and the code below to populate the hash table, I think it would be much easier to use `GVariantDict`. You would initialize one on the stack, call `g_variant_dict_init()` and then call `g_variant_dict_insert()` or `g_variant_dict_insert_value()` to populate it. Then `g_variant_dict_end()` would give you the `a{sv}` variant.

src/eos-crash-metrics.c
```diff
@@ -0,0 +1,181 @@
+/* Copyright 2017 Endless Mobile, Inc. */
```
cosimoc: Please add the license header that the other files in the repo have.

src/eos-crash-metrics.c
smspillaz: Oh handy. Okay, lets use that then.

src/eos-crash-metrics.c
smspillaz: Decided to just make it not-autofree (since we only allocate it at the very end anyway)

src/eos-crash-metrics.c
smspillaz: Done

cosimoc: This function is unused now and should be removed.

src/eos-crash-metrics.c
```diff
@@ -0,0 +1,196 @@
...
+report_crash (const char *binary,
+              gint16 signal,
+              gint64 timestamp,
+              const char *ostree_commit,
+              const char *ostree_url)
+{
+  GVariantDict dict;
+  g_variant_dict_init (&dict, NULL);
+
+  g_variant_dict_insert (&dict, "binary", "v", g_variant_new_string (binary));
```
cosimoc: You either want to use `g_variant_dict_insert_value()` which takes a `GVariant` or use `g_variant_insert(&dict, $key, $type, $value)`

smspillaz: Makes sense

data/90-crash-metrics.conf
```diff
@@ -0,0 +1 @@
+kernel.core_pattern=|/usr/bin/eos-crash-metrics %E %s %t
```
erikos: On my system the "eos-metrics-instrumentation" ends up in "/usr/lib/x86_64-linux-gnu/eos-metrics-instrumentation". I would call the file "90-eos-crash-metrics.conf"

src/eos-crash-metrics.c
```diff
@@ -0,0 +1,174 @@
...
+{
+  const gchar *path = argv[1];
+  const gint16 signal = atoi (argv[2]);
+  const gint64 timestamp = atoll (argv[3]);
+  g_autoptr(OstreeSysroot) sysroot = NULL;
+  g_autoptr(OstreeRepo) repo = NULL;
+  g_autoptr(GError) error = NULL;
+  g_autofree char *url = NULL;
+  g_autofree char *commit = NULL;
+
```
erikos: I think the program should handle the case when we do not have 3 input arguments.

erikos:
I wonder if we should store the path we get from kernel as is e.g. "!app!bin!myprogram" or if we should "convert" it.

data/90-crash-metrics.conf
cosimoc: Yeah, this should be a `.in` file and use `$(libexecdir)/eos-crash-metrics` which will be substituted with the right value at build time.

src/eos-crash-metrics.c
smspillaz: In this case, I think it is fair to assume that we eithet got three arguments or were called incorrectly. I'll add a handler for that

src/eos-crash-metrics.c
```diff
@@ -0,0 +1,174 @@
...
+      return EXIT_FAILURE;
+    }
+
+  if (!url || !commit)
+    {
+      g_warning ("Unable to get OSTree url or commit, perhaps the system has been tampered with?");
+      return EXIT_FAILURE;
+    }
+
+  report_crash (path, signal, timestamp, url, commit);
```
erikos: Just found a good one, the report_crash expects url first, then commit. Will fix it up.

src/eos-crash-metrics.c
```diff
@@ -134,15 +135,25 @@ is_blacklisted_path (const char *path)
 int
 main (int argc, char **argv)
 {
-  const gchar *path = argv[1];
-  const gint16 signal = atoi (argv[2]);
-  const gint64 timestamp = atoll (argv[3]);
+  const gchar *path = NULL;
+  gint16 signal = 0;
+  gint64 timestamp = 0;
```
erikos: Not sure I need to initialize here, as atoi and atoll do default to 0 if the input can not be converted.

src/Makefile.am.inc
```diff
@@ -16,7 +16,10 @@
 ## along with eos-metrics-instrumentation.  If not, see
 ## <http://www.gnu.org/licenses/>.
 
-libexec_PROGRAMS = src/eos-metrics-instrumentation
+libexec_PROGRAMS = \
+	src/eos-metrics-instrumentation \
+	src/eos-crash-metrics \
+    $(NULL)
```
cosimoc: Indentation looks a bit odd here

src/eos-crash-metrics.c
```diff
@@ -0,0 +1,185 @@
...
+  g_autoptr(OstreeSysroot) sysroot = NULL;
+  g_autoptr(OstreeRepo) repo = NULL;
+  g_autoptr(GError) error = NULL;
+  g_autofree char *url = NULL;
+  g_autofree char *commit = NULL;
+
+  if (argc != EXPECTED_NUMBER_ARGS + 1)
+    {
+      g_warning("Not enough arguments passed (%i, expected %i).", argc - 1, EXPECTED_NUMBER_ARGS);
```
cosimoc: Please add a space before paren

src/eos-crash-metrics.c
```diff
@@ -0,0 +1,185 @@
...
+
+  if (is_blacklisted_path (path))
+    {
+      g_message ("%s is blacklisted, not reporting crash", path);
+      return EXIT_SUCCESS;
+    }
+
+  sysroot = load_ostree_sysroot ();
+
+  if (ostree_sysroot_get_repo (sysroot, &repo, NULL, &error))
```
cosimoc: Nitpick: prefer inverting this check and returning early when it fails, so that you can remove the `else` block.

src/eos-crash-metrics.c
```diff
@@ -0,0 +1,183 @@
...
+
+  if (is_blacklisted_path (path))
+    {
+      g_message ("%s is blacklisted, not reporting crash", path);
+      return EXIT_SUCCESS;
+    }
+
+  sysroot = load_ostree_sysroot ();
+
+  if (!ostree_sysroot_get_repo (sysroot, &repo, NULL, &error))
```
cosimoc: `sysroot` can be `NULL` at this point, if `load_ostree_sysroot()` failed.

src/eos-crash-metrics.c
```diff
@@ -0,0 +1,194 @@
...
+  gint64 timestamp = 0;
+  g_autoptr(OstreeSysroot) sysroot = NULL;
+  g_autoptr(OstreeRepo) repo = NULL;
+  g_autoptr(GError) error = NULL;
+  g_autofree char *url = NULL;
+  g_autofree char *commit = NULL;
+
+  if (argc != EXPECTED_NUMBER_ARGS + 1)
+    {
+      g_warning ("Not enough arguments passed (%i, expected %i).", argc - 1, EXPECTED_NUMBER_ARGS);
```
smspillaz: Strictly speaking, this isn't really about "not enough arguments" being passed, but rather that the caller must pass three arguments

smspillaz:
Minor comment, but other than that LGTM

PR #73, opened 2017-09-25
eos-crash-metrics: better information for Flatpak applications

erikos:
https://phabricator.endlessm.com/T19194

### DIFF
--- a/src/eos-crash-metrics.c
+++ b/src/eos-crash-metrics.c
@@ -19,18 +19,49 @@
 
 #include <string.h>
 #include <stdlib.h>
-#include <eosmetrics/eosmetrics.h>
+
+#include <flatpak.h>
 #include <ostree.h>
+
+#include <eosmetrics/eosmetrics.h>
 
 #define PROGRAM_DUMPED_CORE_EVENT "ed57b607-4a56-47f1-b1e4-5dc3e74335ec"
 #define EXPECTED_NUMBER_ARGS 3
+
+typedef struct
+{
+  FlatpakInstalledRef *app;
+  FlatpakInstalledRef *runtime;
+} FlatpakInfo;
+
+static FlatpakInfo *
+flatpak_info_new (FlatpakInstalledRef *app, FlatpakInstalledRef *runtime)
+{
+  FlatpakInfo *info = g_slice_new0 (FlatpakInfo);
+  info->app = g_object_ref (app);
+  info->runtime = g_object_ref (runtime);
+  return info;
+}
+
+static void
+flatpak_info_free (FlatpakInfo *info)
+{
+  g_clear_object (&info->app);
+  g_clear_object (&info->runtime);
+  g_slice_free (FlatpakInfo, info);
+}
+
+G_DEFINE_AUTOPTR_CLEANUP_FUNC (FlatpakInfo, flatpak_info_free)
 
 static void
 report_crash (const char *binary,
               gint16 signal,
               gint64 timestamp,
               const char *ostree_commit,
-              const char *ostree_url)
+              const char *ostree_url,
+              const FlatpakInfo *info,
+              const char *app_url,
+              const char *runtime_url)
 {
   GVariantDict dict;
   g_variant_dict_init (&dict, NULL);
@@ -40,6 +71,20 @@ report_crash (const char *binary,
   g_variant_dict_insert_value (&dict, "timestamp", g_variant_new_int16 (timestamp));
   g_variant_dict_insert_value (&dict, "ostree_commit", g_variant_new_string (ostree_commit));
   g_variant_dict_insert_value (&dict, "ostree_url", g_variant_new_string (ostree_url));
+
+  if (info != NULL)
+    {
+      g_variant_dict_insert_value (&dict, "app_ref",
+                                   g_variant_new_take_string (flatpak_ref_format_ref (FLATPAK_REF (info->app))));
+      g_variant_dict_insert_value (&dict, "app_commit",
+                                   g_variant_new_string (flatpak_ref_get_commit (FLATPAK_REF (info->app))));
+      g_variant_dict_insert_value (&dict, "app_url", g_variant_new_string (app_url));
+      g_variant_dict_insert_value (&dict, "runtime_ref",
+                                   g_variant_new_take_string (flatpak_ref_format_ref (FLATPAK_REF (info->runtime))));
+      g_variant_dict_insert_value (&dict, "runtime_commit",
+                                   g_variant_new_string (flatpak_ref_get_commit (FLATPAK_REF (info->runtime))));
+      g_variant_dict_insert_value (&dict, "runtime_url", g_variant_new_string (runtime_url));
+    }
 
   emtr_event_recorder_record_event_sync (emtr_event_recorder_get_default (),
                                          PROGRAM_DUMPED_CORE_EVENT,
@@ -57,15 +102,16 @@ load_ostree_sysroot (GError **error)
 }
 
 static char *
-get_eos_ostree_repo_url (OstreeRepo *repo)
+get_ostree_repo_url (OstreeRepo *repo, const char *origin)
 {
   GKeyFile *config = NULL;
   g_autoptr(GError) error = NULL;
+  g_autofree char *group = g_strdup_printf ("remote \"%s\"", origin);
   char *url = NULL;
 
   config = ostree_repo_get_config (repo);
 
-  if (!(url = g_key_file_get_string (config, "remote \"eos\"", "url", &error)))
+  if (!(url = g_key_file_get_string (config, group, "url", &error)))
     {
       g_warning ("Unable to read OSTree config for eos remote URL: %s", error->message);
       return NULL;
@@ -137,6 +183,97 @@ normalize_path (char *path)
     path[i] = path[i] == '!' ? '/' : path[i];
 }
 
+static char *
+get_associated_runtime (FlatpakInstalledRef *ref, GError **error)
+{
+  g_autoptr(GBytes) metadata = NULL;
+  gchar *runtime = NULL;
+  g_autoptr(GKeyFile) key_file = NULL;
+
+  metadata = flatpak_installed_ref_load_metadata (ref, NULL, error);
+  if (metadata == NULL)
+    return NULL;
+
+  key_file = g_key_file_new ();
+  if (!g_key_file_load_from_bytes (key_file, metadata, G_KEY_FILE_NONE, error))
+    return NULL;
+
+  if (!(runtime = g_key_file_get_string (key_file, "Application", "runtime", error)))
+    return NULL;
+
+  return runtime;
+}
+
+static FlatpakInfo *
+get_flatpak_info (const char *path,
+                  GError **error)
+{
+  FlatpakInstalledRef *app = NULL;
+  g_autoptr(FlatpakInstalledRef) runtime = NULL;
+  g_autoptr(GPtrArray) xrefs = NULL;
+  guint i;
+  g_autofree char *executable_name = NULL;
+  g_autoptr(FlatpakInstallation) installation = flatpak_installation_new_system (NULL, error);
+
+  if (installation == NULL)
+    return NULL;
+
+  executable_name = g_path_get_basename (path);
+
+  /* we are only interested in apps */
+  xrefs = flatpak_installation_list_installed_refs_by_kind (installation,
+                                                            FLATPAK_REF_KIND_APP,
+                                                            NULL, error);
+  if (xrefs == NULL)
+    return NULL;
+
+  for (i = 0; i < xrefs->len; i++)
+    {
+      g_autofree gchar *executable_path = NULL;
+      FlatpakInstalledRef *xref = g_ptr_array_index (xrefs, i);
+
+      executable_path = g_build_filename (flatpak_installed_ref_get_deploy_dir (xref),
+                                          "files",
+                                          "bin",
+                                          executable_name,
+                                          NULL);
+      /* found a Flatpak with the same application name */
+      if (g_file_test (executable_path, G_FILE_TEST_EXISTS))
+        {
+          app = xref;
+          break;
+        }
+    }
+
+  if (app == NULL)
+    {
+      g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED, "No application with the executable \"%s\" found", executable_name);
+      return NULL;
+    }
+
+  g_autofree char *runtime_name = get_associated_runtime (app, error);
+  if (runtime_name == NULL)
+    return NULL;
+
+  g_auto(GStrv) parts = g_strsplit (runtime_name, "/", 3);
+  if (g_strv_length (parts) != 3)
+    {
+      g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED, "Can not parse runtime name \"%s\"", runtime_name);
+      return NULL;
+    }
+  runtime = flatpak_installation_get_installed_ref (installation,
+                                                    FLATPAK_REF_KIND_RUNTIME,
+                                                    parts[0],
+                                                    parts[1],
+                                                    parts[2],
+                                                    NULL,
+                                                    error);
+  if (runtime == NULL)
+    return NULL;
+
+  return flatpak_info_new (app, runtime);
+}
+
 
 int
 main (int argc, char **argv)
@@ -147,8 +284,11 @@ main (int argc, char **argv)
   g_autoptr(OstreeSysroot) sysroot = NULL;
   g_autoptr(OstreeRepo) repo = NULL;
   g_autoptr(GError) error = NULL;
-  g_autofree char *url = NULL;
-  g_autofree char *commit = NULL;
+  g_autofree char *ostree_url = NULL;
+  g_autofree char *ostree_commit = NULL;
+  g_autoptr(FlatpakInfo) flatpak_info = NULL;
+  g_autofree char *app_url = NULL;
+  g_autofree char *runtime_url = NULL;
 
   if (argc != EXPECTED_NUMBER_ARGS + 1)
     {
@@ -179,16 +319,34 @@ main (int argc, char **argv)
       return EXIT_FAILURE;
     }
 
-  url = get_eos_ostree_repo_url (repo);
-  commit = get_eos_ostree_deployment_commit (sysroot, repo);
+  if (g_str_has_prefix (path, "/app/bin"))
+    {
+      g_message ("%s is likely a Flatpak, get information", path);
+      flatpak_info = get_flatpak_info (path, &error);
+      if (!flatpak_info)
+        {
+          g_warning ("Unable to get flatpak information: %s", error->message);
+          return EXIT_FAILURE;
+        }
+      app_url = get_ostree_repo_url (repo, flatpak_installed_ref_get_origin (flatpak_info->app));
+      runtime_url = get_ostree_repo_url (repo, flatpak_installed_ref_get_origin (flatpak_info->runtime));
+      if (!app_url || !runtime_url)
+        {
+          g_warning ("Unable to get app url or runtime url.");
+          return EXIT_FAILURE;
+        }
+    }
 
-  if (!url || !commit)
+  ostree_url = get_ostree_repo_url (repo, "eos");
+  ostree_commit = get_eos_ostree_deployment_commit (sysroot, repo);
+
+  if (!ostree_url || !ostree_commit)
     {
       g_warning ("Unable to get OSTree url or commit, perhaps the system has been tampered with?");
       return EXIT_FAILURE;
     }
 
-  report_crash (path, signal, timestamp, commit, url);
+  report_crash (path, signal, timestamp, ostree_commit, ostree_url, flatpak_info, app_url, runtime_url);
 
   return EXIT_SUCCESS;
 }
--- a/configure.ac
+++ b/configure.ac
@@ -83,6 +83,7 @@ AC_DEFINE([EOS_INSTRUMENTATION_VERSION], [_EOS_INSTRUMENTATION_VERSION_MACRO], [
 # Required versions of libraries
 # Update these whenever you use a function that requires a certain API version
 EOS_METRICS_REQUIREMENT="eosmetrics-0 >= 0.3.0"
+FLATPAK_REQUIREMENT="flatpak"
 GEOCLUE_REQUIREMENT="geoclue-2.0"
 GLIB_REQUIREMENT="glib-2.0"
 GIO_REQUIREMENT="gio-2.0"
@@ -159,6 +160,7 @@ AC_SUBST([systemdsystemunitdir])
 # ------------------
 PKG_CHECK_MODULES([EOS_INSTRUMENTATION], [
     $EOS_METRICS_REQUIREMENT
+    $FLATPAK_REQUIREMENT
     $GEOCLUE_REQUIREMENT
     $GLIB_REQUIREMENT
     $GIO_REQUIREMENT


src/eos-crash-metrics.c
```diff
@@ -137,6 +140,70 @@ normalize_path (char *path)
...
+                                  const char *executable_name,
+                                  GError **error)
+{
+  g_autoptr(GPtrArray) xrefs = NULL;
+  GPtrArray *applications = NULL;
+  FlatpakInstalledRef *xref = NULL;
+  guint i;
+
+  /* get apps and runtimes */
+	xrefs = flatpak_installation_list_installed_refs (installation,
```
smspillaz: I think this indentation is off

src/eos-crash-metrics.c
```diff
@@ -137,6 +140,70 @@ normalize_path (char *path)
...
+      if (flatpak_ref_get_kind (FLATPAK_REF (xref)) != FLATPAK_REF_KIND_APP)
+        continue;
+
+      executable_path = g_build_filename (flatpak_installed_ref_get_deploy_dir (FLATPAK_INSTALLED_REF (xref)),
+                                          "files",
+                                          "bin",
+                                          executable_name,
+                                          NULL);
+      /* found a Flatpak with the same application name */
+	    if (g_file_test (executable_path, G_FILE_TEST_EXISTS))
```
smspillaz: Indentation off

src/eos-crash-metrics.c
```diff
@@ -137,6 +140,70 @@ normalize_path (char *path)
...
+                                                    NULL, error);
+  if (xrefs == NULL)
+    return NULL;
+
+  applications = g_ptr_array_new ();
+
+  for (i = 0; i < xrefs->len; i++)
+    {
+      g_autofree gchar *executable_path = NULL;
+	    FlatpakInstalledRef *xref = g_ptr_array_index (xrefs, i);
```
smspillaz: Indentation off

src/eos-crash-metrics.c
```diff
@@ -137,6 +140,70 @@ normalize_path (char *path)
...
+
+  executable_name = g_path_get_basename (path);
+
+  applications = get_installed_ref_for_executable (installation, executable_name, error);
+  if (!applications)
+    return NULL;
+
+  /* Even if we find Flatpaks with the same application name
+     we fallback to the first one we find for now */
+	FlatpakInstalledRef *xref = g_ptr_array_index (applications, 0);
```
smspillaz: Indentation off

smspillaz: (In any event, you can just pass the result of `g_ptr_array_index` to `flatpak_ref_get_name` below

src/eos-crash-metrics.c
```diff
@@ -137,6 +140,70 @@ normalize_path (char *path)
...
+  FlatpakInstalledRef *xref = NULL;
+  guint i;
+
+  /* get apps and runtimes */
+	xrefs = flatpak_installation_list_installed_refs (installation,
+                                                    NULL, error);
+  if (xrefs == NULL)
+    return NULL;
+
+  applications = g_ptr_array_new ();
```
smspillaz: Consider using `g_ptr_array_new_full` here to pre-allocate a vector of up to `len` refs. In any event you will need to use `g_ptr_array_new_full` and pass `g_object_unref` to the `GDestroyNotify` parameter to ensure tae the `FlatpakInstalledRef` is unreffed when the returned argument is destroyed.

src/eos-crash-metrics.c
```diff
@@ -181,13 +248,18 @@ main (int argc, char **argv)
...
   if (!url || !commit)
     {
       g_warning ("Unable to get OSTree url or commit, perhaps the system has been tampered with?");
       return EXIT_FAILURE;
     }
 
+  if (g_str_has_prefix (path, "/app/bin"))
+    {
+      g_message ("%s is likely a Flatpak", path);
+      g_warning ("Flatpak name: %s", get_flatpak_application_name (path, &error));
```
smspillaz: I'm not sure if it makes sense to "warn" about this - this is an expected case.

src/eos-crash-metrics.c
```diff
@@ -137,6 +140,70 @@ normalize_path (char *path)
...
+  executable_name = g_path_get_basename (path);
+
+  applications = get_installed_ref_for_executable (installation, executable_name, error);
+  if (!applications)
+    return NULL;
+
+  /* Even if we find Flatpaks with the same application name
+     we fallback to the first one we find for now */
+	FlatpakInstalledRef *xref = g_ptr_array_index (applications, 0);
+  return flatpak_ref_get_name (FLATPAK_REF (xref));
```
smspillaz: If you set the `GDestroyNotify` parameter on the `applications` `GPtrArray` as suggested above, the returned name here outlives the `FlatpakInstalledRef` above, so you will need to transfer it with `g_strdup`

src/eos-crash-metrics.c
```diff
@@ -181,13 +248,18 @@ main (int argc, char **argv)
...
       return EXIT_FAILURE;
     }
 
+  if (g_str_has_prefix (path, "/app/bin"))
+    {
+      g_message ("%s is likely a Flatpak", path);
+      g_warning ("Flatpak name: %s", get_flatpak_application_name (path, &error));
+    }
+
   report_crash (path, signal, timestamp, commit, url);
```
smspillaz: I think as part of this story we also wanted to report the flatpak ref as part of the crash metric, can you pass that through here?

smspillaz: I think we also wanted to figure out which runtimes were linked to the app as well, though that can be done in a followup

src/eos-crash-metrics.c
```diff
@@ -137,6 +145,111 @@ normalize_path (char *path)
...
+        continue;
+
+      executable_path = g_build_filename (flatpak_installed_ref_get_deploy_dir (FLATPAK_INSTALLED_REF (xref)),
+                                          "files",
+                                          "bin",
+                                          executable_name,
+                                          NULL);
+      /* found a Flatpak with the same application name */
+      if (g_file_test (executable_path, G_FILE_TEST_EXISTS))
+        g_ptr_array_add (applications, g_object_ref (xref));
```
cosimoc: I think it would be acceptable to return directly the first ref that matches from this function.

src/eos-crash-metrics.c
```diff
@@ -137,6 +145,111 @@ normalize_path (char *path)
...
+  g_autoptr(GKeyFile) key_file = NULL;
+  const gchar *metadata_data;
+  gsize metadata_len;
+  g_autoptr(GError) local_error = NULL;
+
+  metadata = flatpak_installed_ref_load_metadata (FLATPAK_INSTALLED_REF (ref), NULL, &local_error);
+
+  if (metadata == NULL)
+    {
+      g_warning ("%s", local_error->message);
```
cosimoc: You could have a slightly more explanatory message here (and below)

src/eos-crash-metrics.c
```diff
@@ -137,6 +145,111 @@ normalize_path (char *path)
...
+    }
+
+  metadata_data = g_bytes_get_data (metadata, &metadata_len);
+  key_file = g_key_file_new ();
+
+  if (!g_key_file_load_from_data (key_file, metadata_data, metadata_len, G_KEY_FILE_NONE, &local_error))
+    {
+      if (!g_error_matches (local_error, G_IO_ERROR, G_IO_ERROR_NOT_FOUND))
+        g_warning ("%s", local_error->message);
+        return NULL;
```
cosimoc: I think this should go in the same block as the warning above?

src/eos-crash-metrics.c
```diff
@@ -19,18 +19,45 @@
...
+G_DEFINE_AUTOPTR_CLEANUP_FUNC (FlatpakInfo, flatpak_info_free)
+
 static void
 report_crash (const char *binary,
               gint16 signal,
               gint64 timestamp,
               const char *ostree_commit,
-              const char *ostree_url)
+              const char *ostree_url,
+              const FlatpakInfo * info)
```
smspillaz: space between `*` and `info`

src/eos-crash-metrics.c
```diff
@@ -41,6 +68,22 @@ report_crash (const char *binary,
...
+                                   g_variant_new_string (flatpak_ref_get_name (FLATPAK_REF (info->app))));
+      g_variant_dict_insert_value (&dict, "app_arch",
+                                   g_variant_new_string (flatpak_ref_get_arch (FLATPAK_REF (info->app))));
+      g_variant_dict_insert_value (&dict, "app_branch",
+                                   g_variant_new_string (flatpak_ref_get_branch (FLATPAK_REF (info->app))));
+      g_variant_dict_insert_value (&dict, "app_commit",
+                                   g_variant_new_string (flatpak_ref_get_commit (FLATPAK_REF (info->app))));
+      g_variant_dict_insert_value (&dict, "runtime_name",
+                                   g_variant_new_string (flatpak_ref_get_name (FLATPAK_REF (info->runtime))));
+      g_variant_dict_insert_value (&dict, "runtime_commit",
```
smspillaz: Instead of inserting a bunch of keys like this, it might be better to insert two object-valued parameters, like this:

    "app": {
        "name": ...
    },
    "runtime": {
        "name": ...
    }

src/eos-crash-metrics.c
```diff
@@ -137,6 +180,118 @@ normalize_path (char *path)
...
 }
 
+static inline void
+sanitize_name (gchar *name)
+{
+  gchar *tmp = strchr (name, '/');
+
+  if (tmp != NULL)
+    *tmp = '\0';
+}
```
smspillaz: Instead of this you could use `flatpak_ref_parse` to parse the actual ref

src/eos-crash-metrics.c
```diff
@@ -137,6 +180,118 @@ normalize_path (char *path)
...
+
+  metadata_data = g_bytes_get_data (metadata, &metadata_len);
+  key_file = g_key_file_new ();
+
+  if (!g_key_file_load_from_data (key_file, metadata_data, metadata_len, G_KEY_FILE_NONE, &local_error))
+    {
+      if (!g_error_matches (local_error, G_IO_ERROR, G_IO_ERROR_NOT_FOUND))
+        g_warning ("%s", local_error->message);
+
+      return NULL;
```
smspillaz: Also clear the error here (I know it is marked as `autoptr` above, though the convention appears to be that it should be explicitly cleared so that it is not accidentally reused later.

src/eos-crash-metrics.c
```diff
@@ -137,6 +180,118 @@ normalize_path (char *path)
...
+    }
+
+  if (!(runtime = g_key_file_get_string (key_file, "Application", "runtime", NULL)))
+    return NULL;
```
smspillaz: Bubble up error

src/eos-crash-metrics.c
```diff
@@ -137,6 +180,118 @@ normalize_path (char *path)
...
+      return NULL;
+    }
+
+  metadata_data = g_bytes_get_data (metadata, &metadata_len);
+  key_file = g_key_file_new ();
+
+  if (!g_key_file_load_from_data (key_file, metadata_data, metadata_len, G_KEY_FILE_NONE, &local_error))
+    {
+      if (!g_error_matches (local_error, G_IO_ERROR, G_IO_ERROR_NOT_FOUND))
+        g_warning ("%s", local_error->message);
```
smspillaz: Instead of doing this, you could probably bubble up the error using `g_error_propagate` and then report on failure in the caller if we failed to get the runtime for some reason.

src/eos-crash-metrics.c
smspillaz: Perhaps prefer to bubble up the error

src/eos-crash-metrics.c
```diff
@@ -137,6 +180,118 @@ normalize_path (char *path)
...
+
+  executable_name = g_path_get_basename (path);
+
+  info = flatpak_info_new();
+
+  /* get apps and runtimes */
+  xrefs = flatpak_installation_list_installed_refs (installation,
+                                                    NULL, error);
+  if (xrefs == NULL)
+    return NULL;
```
smspillaz: `info` is leaked here, consider making it `autoptr`

src/eos-crash-metrics.c
```diff
@@ -137,6 +180,118 @@ normalize_path (char *path)
...
+  guint i;
+  g_autofree char *executable_name = NULL;
+  g_autoptr(FlatpakInstallation) installation = flatpak_installation_new_system (NULL, error);
+
+  if (installation == NULL)
+    return NULL;
+
+  executable_name = g_path_get_basename (path);
+
+  info = flatpak_info_new();
```
smspillaz: Space before paren

src/eos-crash-metrics.c
```diff
@@ -137,6 +180,118 @@ normalize_path (char *path)
...
+          continue;
+
+      executable_path = g_build_filename (flatpak_installed_ref_get_deploy_dir (FLATPAK_INSTALLED_REF (xref)),
+                                          "files",
+                                          "bin",
+                                          executable_name,
+                                          NULL);
+      /* found a Flatpak with the same application name */
+      if (g_file_test (executable_path, G_FILE_TEST_EXISTS))
+        info->app = g_object_ref(xref);
```
smspillaz: Space before paren

src/eos-crash-metrics.c
```diff
@@ -137,6 +180,118 @@ normalize_path (char *path)
...
+
+  if (info->app)
+    {
+       g_autofree const char *runtime_name = get_associated_runtime (info->app);
+       for (i = 0; i < runtimes->len; i++)
+         {
+           FlatpakInstalledRef *xref = g_ptr_array_index (runtimes, i);
+           if (g_strcmp0 (flatpak_ref_get_name (FLATPAK_REF (xref)), runtime_name) != 0)
+             continue;
+           info->runtime = g_object_ref(xref);
```
smspillaz: Space before paren

src/eos-crash-metrics.c
```diff
@@ -41,6 +68,22 @@ report_crash (const char *binary,
   g_variant_dict_insert_value (&dict, "ostree_commit", g_variant_new_string (ostree_commit));
   g_variant_dict_insert_value (&dict, "ostree_url", g_variant_new_string (ostree_url));
 
+  if (info->app && info->runtime)
```
smspillaz: `info` is not checked if it is `NULL` here, I think that is what was intended. Though perhaps it may be better to just pass the `FlatpakInstalledRef` instances directly into this function

src/eos-crash-metrics.c
erikos: Doesn't the caller takes the ownership? And that is an "g_autoptr". Fwiw, Valgrind seems to be happy, too.

src/eos-crash-metrics.c
```diff
@@ -137,6 +182,122 @@ normalize_path (char *path)
...
+      if (flatpak_ref_get_kind (FLATPAK_REF (xref)) == FLATPAK_REF_KIND_RUNTIME)
+        {
+          g_ptr_array_add (runtimes, g_object_ref(xref));
+          continue;
+        }
+
+      if ((flatpak_ref_get_kind (FLATPAK_REF (xref)) == FLATPAK_REF_KIND_APP) && info->app)
+          continue;
+
+      executable_path = g_build_filename (flatpak_installed_ref_get_deploy_dir (FLATPAK_INSTALLED_REF (xref)),
```
cosimoc: `xref` is already a `FlatpakInstalledRef` so you should need to cast here

src/eos-crash-metrics.c
```diff
@@ -137,6 +182,122 @@ normalize_path (char *path)
...
+static inline void
+sanitize_name (gchar *name)
+{
+  gchar *tmp = strchr (name, '/');
+
+  if (tmp != NULL)
+    *tmp = '\0';
+}
+
+static const char *
```
cosimoc: Looks like this function returns a new string, so it should be `char *`, not `const char *`

src/eos-crash-metrics.c
```diff
@@ -137,6 +182,122 @@ normalize_path (char *path)
...
+get_associated_runtime (FlatpakInstalledRef *ref, GError **error)
+{
+  g_autoptr(GBytes) metadata = NULL;
+  gchar *runtime = NULL;
+  g_autoptr(GKeyFile) key_file = NULL;
+  const gchar *metadata_data;
+  gsize metadata_len;
+  g_autoptr(GError) local_error = NULL;
+
+  metadata = flatpak_installed_ref_load_metadata (FLATPAK_INSTALLED_REF (ref), NULL, &local_error);
```
cosimoc: `ref` is already a `FlatpakInstalledRef` so you should not need the cast

src/eos-crash-metrics.c
```diff
@@ -137,6 +182,122 @@ normalize_path (char *path)
...
+
+  if (metadata == NULL)
+    {
+      g_propagate_error (error, g_steal_pointer (&local_error));
```
cosimoc: Since you don't do anything with the error locally, it looks like you could simply pass `error` to the above function

src/eos-crash-metrics.c
```diff
@@ -137,6 +182,122 @@ normalize_path (char *path)
...
+      return NULL;
+    }
+
+  metadata_data = g_bytes_get_data (metadata, &metadata_len);
+  key_file = g_key_file_new ();
+
+  if (!g_key_file_load_from_data (key_file, metadata_data, metadata_len, G_KEY_FILE_NONE, &local_error))
```
cosimoc: Use `g_key_file_load_from_bytes()`

src/eos-crash-metrics.c
```diff
@@ -137,6 +182,122 @@ normalize_path (char *path)
...
+    {
+      g_propagate_error (error, g_steal_pointer (&local_error));
```
cosimoc: See the above comment about possibly simplifying this error handling

src/eos-crash-metrics.c
```diff
@@ -137,6 +182,122 @@ normalize_path (char *path)
...
+      return NULL;
+    }
+
+  if (!(runtime = g_key_file_get_string (key_file, "Application", "runtime", &local_error)))
+    {
+      g_propagate_error (error, g_steal_pointer (&local_error));
```
cosimoc: Same here

src/eos-crash-metrics.c
```diff
@@ -137,6 +182,122 @@ normalize_path (char *path)
...
+      /* found a Flatpak with the same application name */
+      if (g_file_test (executable_path, G_FILE_TEST_EXISTS))
+        info->app = g_object_ref (xref);
+    }
+
+  if (info->app)
+    {
+       g_autofree const char *runtime_name = get_associated_runtime (info->app, error);
+       if (runtime_name == NULL)
+         goto end;
```
cosimoc: If you follow my suggestion, I think you  should be able to `return NULL` directly here

src/eos-crash-metrics.c
```diff
@@ -137,6 +182,122 @@ normalize_path (char *path)
...
+       for (i = 0; i < runtimes->len; i++)
```
cosimoc: Instead of having to iterate over all the runtimes here, since you already have the runtime name, you should be able to use e.g. `flatpak_installation_get_installed_ref()` directly.

src/eos-crash-metrics.c
```diff
@@ -179,16 +341,21 @@ main (int argc, char **argv)
...
 
+  if (g_str_has_prefix (path, "/app/bin"))
+    {
+      g_message ("%s is likely a Flatpak, get information", path);
+      flatpak_info = get_flatpak_info (path, &error);
+    }
+
   url = get_eos_ostree_repo_url (repo);
   commit = get_eos_ostree_deployment_commit (sysroot, repo);
-
```
cosimoc: Nitpick: I'd keep this empty line here

src/eos-crash-metrics.c
```diff
@@ -19,18 +19,47 @@
...
+  return info;
+}
+
+static void
+flatpak_info_free (FlatpakInfo *info)
+{
+  if (info->app != NULL)
+    g_object_unref (info->app);
+  if (info->runtime != NULL)
+    g_object_unref (info->runtime);
```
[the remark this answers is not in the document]
erikos: Ahh, excellent much better indeed!

src/eos-crash-metrics.c
erikos: Indeed! That saves me storing the runtimes as well alltogether and simplifies the loop above as well. I can use "flatpak_installation_list_installed_refs_by_kind" and filter for apps.

src/eos-crash-metrics.c
```diff
@@ -137,6 +182,122 @@ normalize_path (char *path)
...
+
+  executable_name = g_path_get_basename (path);
+
+  /* get apps and runtimes */
+  xrefs = flatpak_installation_list_installed_refs (installation,
+                                                    NULL, error);
+  if (xrefs == NULL)
+    return NULL;
+
+  info = flatpak_info_new ();
```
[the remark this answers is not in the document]
erikos: Yes, I simplified the code quite a bit that way. I will double check tomorrow morning, but is quite simpler now. Thanks!

src/eos-crash-metrics.c
```diff
@@ -41,6 +70,22 @@ report_crash (const char *binary,
   g_variant_dict_insert_value (&dict, "ostree_commit", g_variant_new_string (ostree_commit));
   g_variant_dict_insert_value (&dict, "ostree_url", g_variant_new_string (ostree_url));
 
+  if (info != NULL && info->app != NULL && info->runtime != NULL)
+    {
+      g_variant_dict_insert_value (&dict, "app_name",
+                                   g_variant_new_string (flatpak_ref_get_name (FLATPAK_REF (info->app))));
+      g_variant_dict_insert_value (&dict, "app_arch",
+                                   g_variant_new_string (flatpak_ref_get_arch (FLATPAK_REF (info->app))));
+      g_variant_dict_insert_value (&dict, "app_branch",
```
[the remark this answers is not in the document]
erikos: Yes I am sending the refspec now, like: "runtime/com.endlessm.apps.Platform/x86_64/1" and "app/com.endlessm.history.en/x86_64/master". I also send the commit for the app and runtime. 

Also the URL has been requested, is that the location, as in "Location: /var/lib/flatpak/runtime/com.endlessm.apps.Platform/x86_64/1/cb4cc42a6f860bde3fdb53691cb80deca3ff49dea1434535bfe45c8ef3cd5931" or the  origin as in "eos-sdk"?

src/eos-crash-metrics.c
erikos: Excellent, a lot easier, thanks!

src/eos-crash-metrics.c
erikos: Yes, worked.

smspillaz:
Other than that lgtm

erikos:
Thanks for taking another look Sam. We discussed the messaging hierarchy in the standup yesterday, and they want to have a flat message apparently.

smspillaz:
> Thanks for taking another look Sam. We discussed the messaging hierarchy in the standup yesterday, and they want to have a flat message apparently.

If that's the requirement then that makes sense to me. r+ from me.

src/eos-crash-metrics.c
```diff
@@ -179,6 +306,12 @@ main (int argc, char **argv)
       return EXIT_FAILURE;
     }
 
+  if (g_str_has_prefix (path, "/app/bin"))
+    {
+      g_message ("%s is likely a Flatpak, get information", path);
+      flatpak_info = get_flatpak_info (path, &error);
```
cosimoc: There should be some error checking here; if `get_flatpak_info()` fails, I think we should print the error message in a warning, and `return EXIT_FAILURE;`

src/eos-crash-metrics.c
```diff
@@ -137,6 +178,91 @@ normalize_path (char *path)
...
+        {
+          app = xref;
+          break;
+        }
+    }
+
+  if (app == NULL)
+    return NULL;
+
+  g_autofree const char *runtime_name = get_associated_runtime (app, error);
```
cosimoc: This cannot be `const`

src/eos-crash-metrics.c
```diff
@@ -137,6 +178,91 @@ normalize_path (char *path)
...
+  if (runtime_name == NULL)
+    return NULL;
+
+  g_auto(GStrv) parts = g_strsplit (runtime_name, "/", 3);
+  if (g_strv_length (parts) != 3)
+    return NULL;
```
cosimoc: Here you're returning `NULL` without setting the error

src/eos-crash-metrics.c
```diff
@@ -137,6 +178,91 @@ normalize_path (char *path)
...
+
+  FlatpakInstalledRef *runtime = flatpak_installation_get_installed_ref (installation,
```
cosimoc: I believe this should be `g_autoptr()` or will be leaked.

src/eos-crash-metrics.c
```diff
@@ -137,6 +178,91 @@ normalize_path (char *path)
...
+                                                                         FLATPAK_REF_KIND_RUNTIME,
+                                                                         g_strdup (parts[0]),
```
cosimoc: No need to `g_strdup()` these

src/eos-crash-metrics.c
```diff
@@ -41,6 +70,18 @@ report_crash (const char *binary,
   g_variant_dict_insert_value (&dict, "ostree_commit", g_variant_new_string (ostree_commit));
   g_variant_dict_insert_value (&dict, "ostree_url", g_variant_new_string (ostree_url));
 
+  if (info != NULL)
```
cosimoc: If you follow my suggestion below, `info` will never be `NULL` when this function is called.

src/eos-crash-metrics.c
```diff
@@ -41,6 +70,18 @@ report_crash (const char *binary,
...
+    {
+      g_variant_dict_insert_value (&dict, "app_ref",
+                                   g_variant_new_string (flatpak_ref_format_ref (FLATPAK_REF (info->app))));
```
cosimoc: `flatpak_ref_format_ref()` returns a new string, so you should use `g_variant_new_take_string()`

erikos: Oh, thanks good hint, fixed.

src/eos-crash-metrics.c
erikos: It can be null, if we have a non flatpak executable. I left this check in but addressed the failure handling when the flatpak information can not be retrieved.

src/eos-crash-metrics.c
erikos: Yes, and that would segfault with the new error handling code when we try to print out error->message. We actually had another occasion above when no app was found, fixed that one as well and set an error.

src/eos-crash-metrics.c
erikos: Yes done.

PR #75, opened 2018-06-01
Record hardware info 

wjt:
https://phabricator.endlessm.com/T18445

* [ ] ~~graphical hardware IDs~~
* [ ] ~~network hardware IDs~~
* [ ] ~~storage type (HDD/SSD/eMMC/etc)~~
* [x] storage size and usable (by Endless OS) free space
* [x] RAM size
* [x] CPU model (in whatever normal form seems useful - not cores/MHz!)

One could quibble over what to do about the dual-boot case. One might also want to quantize these numbers to the nearest GB, because we don't actually need the exact number of bytes. Thoughts?


wjt:
retest this please!

Pausing here for now. For the remaining 3:

* graphical hardware IDs: what do we want? PCI ID? Human-readable name (in which case we could lift more code from gnome-control-center)? Both?
* network hardware IDs: I guess PCI/USB hardware IDs is what we want. I'm not sure the best way to get them. One could do something like [this logic in eos-phone-home](https://github.com/endlessm/eos-phone-home/blob/98f2a6bbfb80719fc0bd83dc8b4aba8cf8b1c183/eos-phone-home#L279-L297) to enumerate all `/sys/class/net/*` dirs that have a `device`, check the udev subsystem and pull out the PCI ID or USB ID as appropriate. Seems a lot of typing but I have not yet found an appropriate helper library.
* Storage type: the way we seem to determine MMC or not elsewhere is with `lsblk --nodeps --noheading -o SUBSYSTEMS $DEVICE` and checking whether the output starts with `block:mmc`. But how to determine HDD vs SSD? `lsblk` has a `ROTA` attribute which corresponds to `/sys/dev/block/$MAJ:$MIN/queue/rotational` but this attribute is always 1 for USB devices (as far as I can tell) so this does not seem reliable…

wjt:
retest this pleas

wjt:
retest this please

src/eins-hwinfo.c
```diff
@@ -0,0 +1,229 @@
...
+      guint count = GPOINTER_TO_INT (value);
+
+      if (count > G_MAXUINT16)
+        {
+          g_warning ("%s has %u threads; clamping to %" G_GUINT16_FORMAT,
+                     model, count, G_MAXUINT16);
+          count = G_MAXUINT16;
+        }
+
+      g_variant_builder_add (&builder, "{sq}", model, count);
```
wjt: This assumes that `model` is valid UTF-8. The [Wikipedia page for CPUID on x86](https://en.wikipedia.org/wiki/CPUID#EAX=80000002h,80000003h,80000004h:_Processor_Brand_String) says that the processor brand string is ASCII (a subset of UTF-8); and the implementation for ARM is ASCII by construction in the kernel.

However, on ARM 'model name' is just something like "ARMv7 Processor rev 0 (v7l)" which doesn't really tell us what we need…

src/eins-hwinfo.c
```diff
@@ -0,0 +1,119 @@
...
+ * partitions on dual-boot systems, is not reported.
+ *
+ * You might think that given any two of these values for a filesystem, you
+ * could derive the third. That's not the case: typically, 5% of space is
+ * reserved (so used + available = 0.95 * total) but this is a tunable
+ * parameter of the filesystem.
+ */
+#define DISK_SPACE_EVENT "b76c645f-d041-41b0-a4d4-c29f86c18638"
+
+#define RECORD_DISK_SPACE_INTERVAL (60u * 60u * 24u)
```
cosimoc: Would be nice to have the unit in this define as well, and/or a comment that this amounts to a day.

src/eins-hwinfo.c
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+  return g_variant_new ("(ttt)", total, used, available);
+}
+
+static gboolean
+is_mounted (GFile *file)
+{
+  g_autoptr(GFileInfo) info = g_file_query_info (file,
+                                                 G_FILE_ATTRIBUTE_UNIX_IS_MOUNTPOINT,
```
cosimoc: I wonder if this is really the best way to perform this check; `g_file_find_enclosing_mount()` may be more desirable, since I don't know if this would return positively when e.g. the path is an unmounted mountpoint mentioned in `/etc/fstab`

src/eins-hwinfo.c
wjt: Addressed by reimplementing this in terms of `lscpu` which decodes the additional fields necessary to get a more useful model name on ARM.

src/eins-hwinfo.c
wjt: Sure. `G_TIME_SPAN_DAY` exists but is measured in microseconds so I think it's clearer to open-code it here.

wjt: I have also renamed the existing unitless time constant.

src/eins-hwinfo.c
wjt: `G_FILE_ATTRIBUTE_UNIX_IS_MOUNTPOINT` is true if `file`'s parent directory is on a different `st_dev` to `file`. (It's currently `FALSE` for `/` which I think [is a bug](https://gitlab.gnome.org/GNOME/glib/issues/1400).)

`g_file_find_enclosing_mount()` just performs a recursive version of the same check, walking all ancestors until it finds a `st_dev` transition or the root directory. Neither one checks `/etc/fstab`.

ramcq:
LGTM, but one :bike: :derelict_house: - do we want to "bake" the mount structure of split systems into the data type of the event? Given the relative scarcity of split systems, Roddy's mission to obliterate them entirely, and the possibility we may adopt another mount structure in future, perhaps we should just report two events - one SYSROOT_DISK_SPACE_EVENT unconditionally, and one EXTRA_DISK_SPACE_EVENT if present.

I also agree bytes seems an un-necessary level of precision. Didn't we change the installer to MB or GB? We could probably go (uuu) relatively safely then.

ramcq:
For the GPU on x86 at least, I think PCI IDs are sufficient. We can probably iterate /sys/class/drm/cardN/device/{vendor,device}. We should see what that does on our ARMs.

wjt:
> one SYSROOT_DISK_SPACE_EVENT unconditionally, and one EXTRA_DISK_SPACE_EVENT if present

Sure, that's probably a good idea.

> Didn't we change the installer to MB or GB?

Both!

```
13:05 $ git grep 'TrackEvent.*[MG]B"'
src/endless/EndlessUsbToolDlg.cpp:3228: TrackEvent(_T("USBSizeMB"), SelectedDrive.DiskSize / BYTES_IN_MEGABYTE);
src/endless/EndlessUsbToolDlg.cpp:3503: TrackEvent(_T("StorageSizeGB"), approxGB);
src/endless/EndlessUsbToolDlg.cpp:3652: TrackEvent(_T("MaximumInstallSizeGB"), ROUND_TO_NEAREST_GIGABYTE(maximumInstallSize));
```

wjt:
Updated to round sizes to the nearest decimal gigabyte, and split the disk event in two.

I am prepared to be convinced that we should be using gibibytes because that's how the Winstaller rounds (for better or for worse) but, eh…

wjt:
Assuming CR pass, I'll squash away all the fixup commits before merging.

src/eins-hwinfo.c
```diff
@@ -130,30 +149,51 @@ record_disk_space (gpointer data)
...
-  GVariant *payload = g_variant_new ("(@(ttt)@(ttt))", root_payload, extra_payload);
+  g_autoptr(GError) error = NULL;
+  GVariant *root_payload =
+    eins_hwinfo_get_disk_space_for_partition (root, &error);
+
+  if (error != NULL)
+    {
+      g_warning ("Couldn't get disk space for /: %s", error->message);
+      g_clear_error (&error);
+    }
```
ramcq: Are you intentionally falling through here so as to return (0, 0, 0) in the error case?

src/eins-hwinfo.c
```diff
@@ -0,0 +1,417 @@
...
+ */
+#define EXTRA_DISK_SPACE_EVENT "da505554-4248-4a38-bb32-84ab58e45a6d"
+
+/* One day */
+#define RECORD_DISK_SPACE_INTERVAL_SECONDS (60u * 60u * 24u)
+
+#define ONE_GB_IN_BYTES (G_GUINT64_CONSTANT (1000 * 1000 * 1000))
+
+/*
+ * The amount of physical memory accessible to Endless OS, in bytes. Reported
```
ramcq: Inconsistent with the code.

src/eins-hwinfo.c
```diff
@@ -0,0 +1,417 @@
...
+  return GPOINTER_TO_INT (data);
+}
+
+GVariant *
+eins_hwinfo_get_ram_size (void)
+{
+  glibtop_mem mem;
+
+  glibtop_get_mem (&mem);
+  return g_variant_new_uint32 (round_to_nearest_gigabyte (mem.total));
```
ramcq: I... er... prefer RAM in megabytes? Maybe even mibibytes. Things like UMA apertures and other BIOS options can steal non-integer number of gigabytes from the system.

src/eins-hwinfo.c
wjt: I am, but I could be convinced to just not submit anything in the error case.

src/eins-hwinfo.c
wjt: good point!

src/eins-hwinfo.c
wjt: also a good point!

src/eins-hwinfo.c
ramcq: If (0, 0, 0) is to be reported and semantically meaningful, it should be documented - but the inclusion of such error values within the number space means we can't meaningfully rely on any aggregate operations like minimum, median, mode or mean without filtering the values first. I favor not submitting this value, because if we wished to count the occurrence of errors, we can infer the "NULL" value by whether the event is reported and compare to the number of reporting systems.

wjt: It was documented in the ((ttt)(ttt)) version of this event but I lost it in the split. Anyway, your argument is convincing.

src/eins-hwinfo.c
wjt: Changed both to power-of-2; changed this to mebibytes. There's an argument that we should always round towards 0. I implemented round-to-nearest because that's what I implemented in the winstaller, but can't remember why that's what I implemented there.

PR #81, opened 2019-07-26
Remove city location metrics

ptomato:
https://phabricator.endlessm.com/T27349

### DIFF
--- a/src/eos-metrics-instrumentation.c
+++ b/src/eos-metrics-instrumentation.c
@@ -364,12 +364,8 @@ get_image_version (void)
 }
 
 static gboolean
-record_image_version (gpointer unused)
+record_image_version (const char *image_version)
 {
-  g_autofree gchar *image_version = NULL;
-
-  image_version = get_image_version ();
-
   if (image_version != NULL)
     emtr_event_recorder_record_event (emtr_event_recorder_get_default (),
                                       EOS_IMAGE_VERSION_EVENT,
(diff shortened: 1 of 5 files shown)


wjt:
I aborted the CI build after 2 days without progress.

retest this please!

ptomato:
Failed due to filesystem busy error.

bot, retest this please

ptomato:
It was my bad that the test was hanging, I had forgotten that the tests were nontrivial...

Review appreciated on the way I fixed the test, otherwise I will go ahead and merge this tomorrow.

src/eos-metrics-instrumentation.c
```diff
@@ -948,7 +953,7 @@ record_network_change (GDBusProxy *dbus_proxy,
       /* schedule recording the network ID provided we have a default route */
       if (new_network_state >= NM_STATE_CONNECTED_SITE)
         {
-          g_idle_add ((GSourceFunc) record_network_id, GINT_TO_POINTER (FALSE));
+          g_idle_add ((GSourceFunc) record_network_id, NULL);
```
wjt: Seems wrong – this will pass `NULL` as `image_version` so will never do anything, even on solutions images.

tests/test-location.c
```diff
@@ -0,0 +1,44 @@
...
+  return G_SOURCE_REMOVE;
+}
+
+gint
+main (gint                argc,
+      const gchar * const argv[])
+{
+  GMainLoop *main_loop = g_main_loop_new (NULL, TRUE);
+  g_idle_add ((GSourceFunc) record_location_metric, "solutions-fake-image");
+  g_unix_signal_add (SIGTERM, (GSourceFunc) quit_main_loop, main_loop);
```
wjt: I guess this is here because there is otherwise a race between this mock daemon closing itself and exiting cleanly, and the Python test calling `.terminate()` to kill it. But the test doesn't check the exit status of the mock daemon so is this needed?

ptomato: That's true, there's no reason why this needs to exit cleanly.

src/eos-metrics-instrumentation.c
ptomato: ah, this was a bit careless. Fixed.

ptomato:
I added another commit to fix the test timeout, as well.

PR #82, opened 2019-09-10
location: Hardcode fnde and impact images as Solutions images

ptomato:
We only record location on Solutions images and this category should
include image names starting with 'impact' and 'fnde'. Hardcoding for
the time being until we find a better solution.

https://phabricator.endlessm.com/T27743


treitter:
Note that the comment above the modified code is now out-of-date

PR #83, opened 2019-09-16
network ID: Hardcode fnde and impact images as Solutions images

ptomato:
We only record the network ID on Solutions images and this category
should include image names starting with 'impact' and 'fnde'. Hardcoding
for the time being until we find a better solution.

https://phabricator.endlessm.com/T27819

### DIFF
--- a/src/eos-metrics-instrumentation.c
+++ b/src/eos-metrics-instrumentation.c
@@ -892,13 +892,13 @@ record_network_id_impl (const char *image_version,
 {
   guint32 network_id;
 
-  /* Network ID is only needed for analysis on Solutions images */
+  /* Network ID is only needed for analysis on certain partner images */
   if (!image_version ||
       !(g_str_has_prefix (image_version, "fnde-") ||
         g_str_has_prefix (image_version, "impact-") ||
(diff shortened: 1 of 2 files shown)


src/eos-metrics-instrumentation.c
```diff
@@ -893,7 +893,10 @@ record_network_id_impl (const char *image_version,
   guint32 network_id;
 
   /* Network ID is only needed for analysis on Solutions images */
```
treitter: Nitpick: this comment is now out-of-date. Should be "on select images" or similar to avoid aliasing the code below.

ptomato: Basically all of these are Solutions deals, no? In any case I added a commit rewording the comments and debug messages.

treitter: The `fnde` images are for Solutions deals but, pedantically, the `impact` images are more Philanthropy (with the work done by Solutions). In theory, if/when Philanthropy has engineers, they'll do the work and manage the images themselves.

And it might be helpful to clarify the distinction between `solutions` (the image product) and Solutions the company.

PR #86, opened 2019-10-22
Stop sending NETWORK_STATUS_CHANGED_EVENT

ramcq:
This event (5fae6179-e108-4962-83be-c909259c0584) is the most
frequently-submitted event, with 500 million submissions in a year.
It's not particularly valuable, given that we have never had cause nor
desire to try and analyse the data in the 4 years since I joined
Endless. Stop sending it to save space and bandwidth, as the new metrics
system will also ignore and not store it.

https://phabricator.endlessm.com/T28301


ramcq:
@wjt Warning - uncompiled and untested bilge. :)

PR #88, opened 2020-02-12
Don't include literal quotes in OS_VERSION_EVENT

wjt:
Previously, we would report this event as (on my system):

    ('"Endless"', '"3.8.0"', '')

As can be seen by the code comment, this was done for simplicity when
parsing /etc/os-release. (Personally I don't really see that it's
simpler to parse it on the server side, but there we go.)

Nonetheless, GLib 2.63.1 added API for retrieving parsed values from
/etc/os-release. Use it.

TODO:

- [x] Should this get a new UUID? If so, we could drop the unused fields.

https://phabricator.endlessm.com/T29322


wjt:
This is a draft because I want to ensure Azafea is modified to drop the quotes from events sent on old systems before we start submitting data without them. Otherwise, querying this will be even harder.

wjt:
We decided that we shouldn't use a new UUID. Quoth @bochecha on https://phabricator.endlessm.com/T29322 –

> If you changed the UUID and made a new event though, then we'd have 2 event models in Azafea. We could make them point to the same underlying table though, it would just require the appropriate comments to clear the confusion of future code readers. It might confuse the migration system (Alembic) as well, I'm not sure about that. It's something I'd need to explore if we go that route, but I'm not too worried.

PR #89, opened 2020-02-21
Improve coredump data

wjt:
This adds the machine architecture and human-readable OS version to the information submitted about core dumps, and also correctly identifies the booted version of the OS.

https://phabricator.endlessm.com/T29404

### DIFF
--- a/src/eos-crash-metrics.c
+++ b/src/eos-crash-metrics.c
@@ -1,4 +1,4 @@
-/* Copyright 2017 Endless Mobile, Inc.
+/* Copyright © 2017–2020 Endless Mobile, Inc.
  *
  * This file is part of eos-metrics-instrumentation.
  *
@@ -124,10 +124,7 @@ static char *
 get_eos_ostree_deployment_commit (OstreeSysroot *sysroot, OstreeRepo *repo)
 {
   OstreeDeployment *deployment = ostree_sysroot_get_booted_deployment (sysroot);
-  GKeyFile *config = NULL;
-  g_autoptr(GError) error = NULL;
-  g_autofree char *refspec = NULL;
-  char *commit = NULL;
+  const char *csum = NULL;
 
   if (!deployment)
     {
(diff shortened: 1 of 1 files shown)


src/eos-crash-metrics.c
```diff
@@ -120,21 +123,40 @@ get_ostree_repo_url (OstreeRepo *repo, const char *origin)
...
+   */
+  if (ostree_repo_load_variant (repo, OSTREE_OBJECT_TYPE_COMMIT, csum,
+                                &commit, NULL))
+    commit_metadata = g_variant_get_child_value (commit, 0);
+
+  if (commit_metadata)
+    (void) g_variant_lookup (commit_metadata, OSTREE_COMMIT_META_KEY_VERSION, "&s", &version);
+
+  *commit_out = g_strdup (csum);
+  *version_out = g_strdup (version);
```
pwithnall: Nitpick: Would be good to either `g_return_val_if_fail (commit_out != NULL, FALSE)` for each of these, or `if (commit_out != NULL) *commit_out = g_strdup (…)`

src/eos-crash-metrics.c
```diff
@@ -120,21 +123,40 @@ get_ostree_repo_url (OstreeRepo *repo, const char *origin)
...
   const char *csum = NULL;
+  g_autoptr(GVariant) commit = NULL;
+  g_autoptr(GVariant) commit_metadata = NULL;
+  const char *version = NULL;
 
   if (!deployment)
     {
       g_warning ("OSTree deployment is not currently booted, cannot read state");
-      return NULL;
+      return FALSE;
```
pwithnall: Nitpick: Slightly safer to clear both out-variables to `NULL` if returning `FALSE`. Or (less safe but fine) document that their values are undefined if `FALSE` is returned.

src/eos-crash-metrics.c
```diff
@@ -345,6 +350,12 @@ main (int argc, char **argv)
         }
     }
 
+  if (uname(&name) == -1)
```
pwithnall:
```suggestion
  if (uname (&name) < 0)
```
(sidesteps the problem of “but what if it somehow returns something `<-1`??”)

src/eos-crash-metrics.c
wjt: Addressed by adding a `g_return_val_if_fail` guard.

src/eos-crash-metrics.c
wjt: Addressed by requiring that they are initialized to `NULL` by the caller.

wjt:
Addressed those points and also removed an unneeded `GError` variable.

PR #90, opened 2020-05-14
Ignore empty location label keys

wjt:
If you use eos-label-location and don't give an answer to one or more of
its questions, you get a file like this:

    [Label]
    facility = Will's House
    city = London
    state =

The 'site' page in our branch of gnome-initial-setup similarly writes
empty files like this:

    [Label]
    id =
    city =
    state =
    street =
    country =
    facility =

As a result of this second case, over 99% of the rows in our production
table for this event are:

    {"id": "", "city": "", "state": "", "street": "", "country": "", "facility": ""}

Handle the first case by ignoring empty fields. Handle the second case
by not reporting a metric if there are no non-empty fields. Test these
cases.

https://phabricator.endlessm.com/T30469

### DIFF
--- a/tests/test-location-label.c
+++ b/tests/test-location-label.c
@@ -23,7 +23,7 @@ static void
 test_empty_keyfile (void)
 {
   g_autoptr(GKeyFile) kf = g_key_file_new ();
-  g_autoptr(GVariant) payload = build_location_label_event (kf);
+  GVariant *payload = build_location_label_event (kf);
 
   g_assert_null (payload);
 }
@@ -35,7 +35,19 @@ test_empty_group (void)
   gboolean ret = g_key_file_load_from_data (kf, "[Label]\n", -1, G_KEY_FILE_NONE, NULL);
   g_assert_true (ret);
 
-  g_autoptr(GVariant) payload = build_location_label_event (kf);
+  GVariant *payload = build_location_label_event (kf);
+
+  g_assert_null (payload);
+}
+
+static void
+test_only_blank_keys (void)
+{
+  g_autoptr(GKeyFile) kf = g_key_file_new ();
+  g_key_file_set_string (kf, "Label", "id", "");
+
+  GVariant *payload = build_location_label_event (kf);
+
   g_assert_null (payload);
 }
 
@@ -46,7 +58,7 @@ test_only_populated_keys (void)
   g_key_file_set_string (kf, "Label", "facility", "Aperture Science");
   g_key_file_set_string (kf, "Label", "city", "Unknown");
 
-  g_autoptr(GVariant) payload = build_location_label_event (kf);
+  g_autoptr(GVariant) payload = g_variant_take_ref (build_location_label_event (kf));
   g_assert_nonnull (payload);
 
   g_assert_cmpuint (g_variant_n_children (payload), ==, 2);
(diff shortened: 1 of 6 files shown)


tests/test-location-label.c
```diff
@@ -0,0 +1,77 @@
...
+  g_autoptr(GVariant) payload = build_location_label_event (kf);
+
+  g_assert_null (payload);
+}
+
+static void
+test_empty_group (void)
+{
+  g_autoptr(GKeyFile) kf = g_key_file_new ();
+  gboolean ret = g_key_file_load_from_data (kf, "[Label]\n", -1, G_KEY_FILE_NONE, NULL);
```
mwleeds, line 35: nit: would be a bit better to move `#define LOCATION_LABEL_GROUP "Label"` to the header file and use that.

tests/test-location-label.c
```diff
@@ -0,0 +1,77 @@
...
+ * <http://www.gnu.org/licenses/>.
+ */
+
+#include "eins-location-label.h"
+
+static void
+test_empty_keyfile (void)
+{
+  g_autoptr(GKeyFile) kf = g_key_file_new ();
+  g_autoptr(GVariant) payload = build_location_label_event (kf);
```
mwleeds, line 26: It's a floating GVariant so it doesn't make sense to use `g_autoptr` right?

mwleeds: (and similarly for the other tests)

mwleeds:
> TODO:
> 
> Should we actually be submitting {} if there are no non-empty
> fields, to handle the case where a machine previously had location
> label information and now does not?

That seems like it would be exceedingly rare in practice, but you're right perhaps we should handle it.

> How about if the file doesn't exist at all?

We already don't report a metric in that case. And if you mean that it existed and was deleted, that should probably be handled the same as the above case (no non-empty fields).

tests/test-location-label.c
wjt: I'm not sure this line would be any clearer or more maintainable if it were written as:

```suggestion
  gboolean ret = g_key_file_load_from_data (kf,  "[" LOCATION_LABEL_GROUP "]\n", -1, G_KEY_FILE_NONE, NULL);
```

or as:

```suggestion
  g_autofree gchar *data = g_strdup_printf ("[%s]\n", LOCATION_LABEL_GROUP);
  gboolean ret = g_key_file_load_from_data (kf, "[Label]\n", -1, G_KEY_FILE_NONE, NULL);
```

tests/test-location-label.c
wjt: It's floating, but, it still has a ref count of 1, and unreffing it will still release it. (As far as I understand from the documentation and the source.)

wjt:
The thing about submitting an empty event in the case where the file doesn't exist, or where it exists but is empty, is that the reason I was looking at this in the first place was the millions of completely useless rows in the database, and submitting an empty event in either of those cases would mean those would still be around unless the server took care to deduplicate them.

Which… maybe it should. Other options might include:

* Send an empty event //when the file is deleted// since this code already monitors for changes, but not on startup
* Do nothing since this case is extremely rare

I am tempted by the latter.

tests/test-location-label.c
wjt: I adjusted the tests to take the floating reference where appropriate, and drop `g_autoptr` in the "it's NULL" cases.

tests/test-location-label.c
wjt: On reflection I actively prefer for the test to inline the constant.

Consider what would happen if the test used `LOCATION_LABEL_GROUP` as you suggest, and then someone changed the definition of that constant. The test would still pass – but the application would no longer do the right thing, since it would no longer read the file as it exists on machines in the wild.

So unless you feel strongly otherwise I'd prefer to keep it inlined.

wjt:
I've updated the commit message to explain that it's not handling the case where the file becomes empty, with a brief rationale.

tests/test-location-label.c
mwleeds: Touché

PR #91, opened 2021-05-25
Remove unneeded events from eos-metrics-instrumentation

starnight:
* Windows license tables event
* Disk space on extra partition event
* (Geo)location event
* Uptime event
* Network ID event

https://phabricator.endlessm.com/T31354


dbnicholson:
@liZe I don't know if you're familiar with the events that we'd like to remove, but if so could you have a look at this PR while @wjt is out this week?

liZe:
> @liZe I don't know if you're familiar with the events that we'd like to remove, but if so could you have a look at this PR while @wjt is out this week?

OK! I’ll review this issue on Thursday.

wjt (approved):

PR #93, opened 2021-07-26
hwinfo: Implement combined computer hardware info event

starnight:
Introduce new COMPUTER_HWINFO_EVENT to replace/merge RAM_SIZE_EVENT,
SYSROOT_DISK_SPACE_EVENT and CPU_MODELS_EVENT. Modified corresponding
test plans as well.

https://phabricator.endlessm.com/T32205

### DIFF
--- a/src/eins-hwinfo.c
+++ b/src/eins-hwinfo.c
@@ -24,11 +24,36 @@
 #include <json-glib/json-glib.h>
 
 /*
- * Reported at system startup, and every RECORD_DISK_SPACE_INTERVAL_SECONDS
- * after startup. The payload is a triple of uint32, (uuu), representing the
- * the total size, space used, and space available on the root filesystem,
- * measured in gibibytes (2^30 bytes). We round to the nearest gibibyte: we
- * have no need of a precise figure in the reported data.
+ * Computer hardware information event with payload "(uuuua(sqd))".
+ *
+ * Field  | Description
+ * -------+----------------------------------
+ *      u | Please see RAM section
+ *    uuu | Please see Root partition section
+ * a(sqd) | Please see CPU section
+ */
+
+#define COMPUTER_HWINFO_EVENT "81f303aa-448d-443d-97f9-8d8a9169321c"
+
+/* 24 hours */
+#define RECORD_COMPUTER_HWINFO_INTERVAL_SECONDS (60u * 60u * 24u)
+
+/*
+ * RAM:
+ * The amount of physical memory accessible to Endless OS, in mebibytes (2^20
+ * bytes). The payload is a uint32 (u).
+ */
+
+#define ONE_MIB_IN_BYTES (G_GUINT64_CONSTANT (1024 * 1024))
+
+#define RAMINFO_TYPE_STRING "u"
+
+/*
+ * Root partition:
+ * The payload is a triple of uint32, (uuu), representing the the total size,
+ * space used, and space available on the root filesystem, measured in gibibytes
+ * (2^30 bytes). We round to the nearest gibibyte: we have no need of a precise
+ * figure in the reported data.
  *
  * On dual-boot installations, this refers to the Endless OS image file, not
  * the Windows partition it is hosted on.
@@ -40,36 +65,16 @@
  * could derive the third. That's not the case: typically, 5% of space is
  * reserved (so used + available = 0.95 * total) but this is a tunable
  * parameter of the filesystem.
- *
- * If disk space cannot be determined for some reason, this event will not be
- * reported.
  */
-#define SYSROOT_DISK_SPACE_EVENT "5f58024f-3b99-47d3-a17f-1ec876acd97e"
-
-/* One day */
-#define RECORD_DISK_SPACE_INTERVAL_SECONDS (60u * 60u * 24u)
-
-/* The presence of this file indicates that the first-boot resize of the root
- * filesystem is complete.
- *
- * https://github.com/endlessm/eos-boot-helper/blob/master/eos-firstboot
- */
-#define BOOTED_FLAG_FILE_PATH "/var/eos-booted"
 
 #define ONE_GIB_IN_BYTES (G_GUINT64_CONSTANT (1024 * 1024 * 1024))
 
-/*
- * The amount of physical memory accessible to Endless OS, in mebibytes (2^20
- * bytes). Reported once at system startup. The payload is a uint32 (u).
- */
-#define RAM_SIZE_EVENT "aee94585-07a2-4483-a090-25abda650b12"
-
-#define ONE_MIB_IN_BYTES (G_GUINT64_CONSTANT (1024 * 1024))
+#define ROOTFS_SPACE_TYPE_STRING "uuu"
 
 /*
- * CPUs in the system. Reported once at system startup. The payload is an array
- * of triples -- a(sqd) -- containing the following information for each group
- * of similar cores/threads:
+ * CPU:
+ * CPUs in the system. The payload is an array of triples -- a(sqd) --
+ * containing the following information for each group of similar cores/threads:
  *
  * Field | Type   | Description              | Default if unknown
  * ------+--------+--------------------------+-------------------
@@ -91,7 +96,12 @@
  * array, containing details of the big and LITTLE cores. In practice, the
  * current implementation only reports the currently-active cores.
  */
-#define CPU_MODELS_EVENT "4a75488a-0d9a-4c38-8556-148f500edaf0"
+
+#define CPUINFO_TYPE_STRING "(sqd)"
+#define CPUINFO_ARRAY_TYPE_STRING "a" CPUINFO_TYPE_STRING
+
+#define COMPUTER_HWINFO_TYPE_STRING "(" RAMINFO_TYPE_STRING \
+  ROOTFS_SPACE_TYPE_STRING "@" CPUINFO_ARRAY_TYPE_STRING ")"
 
 static guint32
 round_to_nearest (guint64 size,
@@ -105,14 +115,32 @@ round_to_nearest (guint64 size,
   return (guint32) CLAMP (size / divisor, 0, G_MAXUINT32);
 }
 
-GVariant *
-eins_hwinfo_get_disk_space_for_partition (GFile   *file,
-                                          GError **error)
+guint32
+eins_hwinfo_get_ram_size (void)
+{
+  glibtop_mem mem;
+  guint32 size_mib;
+
+  glibtop_get_mem (&mem);
+  size_mib = round_to_nearest (mem.total, ONE_MIB_IN_BYTES);
+  return size_mib;
+}
+
+gboolean
+eins_hwinfo_get_disk_space_for_partition (GFile          *file,
+                                          DiskSpaceType  *diskspace,
+                                          GError        **error)
 {
   g_autoptr(GFileInfo) info = NULL;
   guint64 total = 0;
   guint64 used = 0;
-  guint64 available = 0;
+  guint64 free = 0;
+
+  g_return_val_if_fail (diskspace != NULL, FALSE);
+
+  diskspace->total = 0;
+  diskspace->used = 0;
+  diskspace->free = 0;
 
   info = g_file_query_filesystem_info (file,
                                        G_FILE_ATTRIBUTE_FILESYSTEM_SIZE ","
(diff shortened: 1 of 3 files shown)


src/eins-hwinfo.c
```diff
@@ -24,11 +24,36 @@
...
 
 /*
- * Reported at system startup, and every RECORD_DISK_SPACE_INTERVAL_SECONDS
- * after startup. The payload is a triple of uint32, (uuu), representing the
- * the total size, space used, and space available on the root filesystem,
- * measured in gibibytes (2^30 bytes). We round to the nearest gibibyte: we
- * have no need of a precise figure in the reported data.
+ * Computer hardware information event with payload "(uuuua(sqd))".
+ *
+ * Feild  | Description
```
wjt, line 29:
```suggestion
 * Field  | Description
```

src/eins-hwinfo.c
```diff
@@ -24,11 +24,36 @@
...
+
+#define COMPUTER_HWINFO_EVENT "81f303aa-448d-443d-97f9-8d8a9169321c"
+
+/* 24 hours */
+#define RECORD_COMPUTER_HWINFO_INTERVAL_SECONDS (60u * 60u * 24u)
+
+/*
+ * RAM:
+ * The amount of physical memory accessible to Endless OS, in mebibytes (2^20
+ * bytes). Reported once at system startup. The payload is a uint32 (u).
```
wjt, line 44:
```suggestion
 * bytes). The payload is a uint32 (u).
```

src/eins-hwinfo.c
```diff
@@ -40,33 +65,14 @@
...
- * bytes). Reported once at system startup. The payload is a uint32 (u).
- */
-#define RAM_SIZE_EVENT "aee94585-07a2-4483-a090-25abda650b12"
-
-#define ONE_MIB_IN_BYTES (G_GUINT64_CONSTANT (1024 * 1024))
+#define ROOTFS_SPACE_TYPE_STRING "uuu"
 
 /*
+ * CPU:
  * CPUs in the system. Reported once at system startup. The payload is an array
```
wjt, line 76:
```suggestion
 * CPUs in the system. The payload is an array
```

src/eins-hwinfo.c
```diff
@@ -105,9 +115,21 @@ round_to_nearest (guint64 size,
...
+{
+  glibtop_mem mem;
+  guint32 size_mib;
+
+  glibtop_get_mem (&mem);
+  size_mib = round_to_nearest (mem.total, ONE_MIB_IN_BYTES);
+  return size_mib;
+}
+
+void
```
wjt, line 129: Conventionally this function would return `TRUE` (with `diskspace` set and `error` untouched) or `FALSE` (with `diskspace` untouched and `error` set).

```suggestion
gboolean
```

and then adjust callers to use

```
if (eins_hwinfo_get_disk_space_for_partition (...))
  {
    // success path
  }
else
  {
    // error path
  }
```

The other option is to log the error within this function, initialise `diskspace` to (0, 0, 0), and remove the `error` parameter. This might be better since it would allow the rest of the event to be sent.

src/eins-hwinfo.c
```diff
@@ -105,9 +115,21 @@ round_to_nearest (guint64 size,
...
+eins_hwinfo_get_disk_space_for_partition (GFile          *file,
+                                          DiskSpaceType  *diskspace,
+                                          GError        **error)
 {
```
wjt, line 133:
```suggestion
{
g_return_val_if_fail (diskspace != NULL, FALSE);
```

(but moved below the variable declarations)

src/eins-hwinfo.c
```diff
@@ -122,10 +144,9 @@ eins_hwinfo_get_disk_space_for_partition (GFile   *file,
                                        error);
   if (info == NULL)
     {
-      /* TODO: g_file_peek_path() from GLib 2.56. */
-      g_autofree gchar *path = g_file_get_path (file);
+      const gchar *path = g_file_peek_path (file);
       g_prefix_error (error, "Couldn't get disk space for %s: ", path);
-      return NULL;
+      return;
```
wjt, line 149:
```suggestion
      return TRUE;
```

src/eins-hwinfo.c
```diff
@@ -135,129 +156,18 @@ eins_hwinfo_get_disk_space_for_partition (GFile   *file,
...
-  g_autoptr(GFileInfo) info = g_file_query_info (file,
-                                                 G_FILE_ATTRIBUTE_UNIX_IS_MOUNTPOINT,
-                                                 G_FILE_QUERY_INFO_NOFOLLOW_SYMLINKS,
-                                                 NULL, NULL);
-
-  return info != NULL &&
-    g_file_info_get_attribute_boolean (info, G_FILE_ATTRIBUTE_UNIX_IS_MOUNTPOINT);
+  diskspace->total = round_to_nearest (total, ONE_GIB_IN_BYTES);
+  diskspace->used = round_to_nearest (used, ONE_GIB_IN_BYTES);
+  diskspace->free = round_to_nearest (available, ONE_GIB_IN_BYTES);
```
wjt, line 161:
```suggestion
  diskspace->free = round_to_nearest (available, ONE_GIB_IN_BYTES);
  
  return TRUE;
```

src/eins-hwinfo.c
```diff
@@ -135,129 +156,18 @@ eins_hwinfo_get_disk_space_for_partition (GFile   *file,
...
-}
-
-static gboolean
-record_ram_size (gpointer unused)
-{
-  emtr_event_recorder_record_event (emtr_event_recorder_get_default (),
-                                    RAM_SIZE_EVENT,
-                                    eins_hwinfo_get_ram_size ());
-  return G_SOURCE_REMOVE;
+  eins_hwinfo_get_disk_space_for_partition (root, diskspace, &error);
```
wjt, line 170: If you are just going to ignore the error you can just pass `NULL`. But the caller should log the error with `g_warning()` as it previously did.

src/eins-hwinfo.c
```diff
@@ -454,22 +364,107 @@ eins_hwinfo_get_cpu_info (void)
   return eins_hwinfo_parse_lscpu_json (json_data, (gssize) json_size);
 }
 
+GVariant *
+eins_hwinfo_get_computer_hwinfo (void)
+{
+  guint32 ramsize = eins_hwinfo_get_ram_size ();
+  DiskSpaceType diskspace = {};
+  GVariant *cpuinfo = eins_hwinfo_get_cpu_info();
```
wjt, line 372: This function can fail and return `NULL`. Passing `NULL` to `g_variant_new()` below will crash. This function should either gracefully return `NULL` in that case, or gracefully pass a GVariant  of an empty `a(squ)` array to `g_variant_new()` below. I think passing a new empty array is better since then we can at least report the ram size & disk space.

tests/test-hwinfo.c
```diff
@@ -270,5 +297,7 @@ main (int   argc,
 
   g_test_add_func ("/hwinfo/cpu/current", test_get_cpu_info_for_current_system);
```
wjt, line 298: Isn't this test now redundant with `test_get_computer_hwinfo` which makes the same assertions?

src/eins-hwinfo.c
```diff
@@ -91,7 +97,11 @@
  * array, containing details of the big and LITTLE cores. In practice, the
  * current implementation only reports the currently-active cores.
  */
-#define CPU_MODELS_EVENT "4a75488a-0d9a-4c38-8556-148f500edaf0"
+
+#define CPUINFO_TYPE_STRING "(sqd)"
+
+#define COMPUTER_HWINFO_TYPE_STRING "(" RAMINFO_TYPE_STRING \
+  ROOTFS_SPACE_TYPE_STRING "@a" CPUINFO_TYPE_STRING ")"
```
wjt, line 104:
```suggestion
#define CPUINFO_TYPE_STRING "a(sqd)"

#define COMPUTER_HWINFO_TYPE_STRING "(" RAMINFO_TYPE_STRING \
  ROOTFS_SPACE_TYPE_STRING "@" CPUINFO_TYPE_STRING ")"
```

wjt (changes requested):
As written, this event will be sent once at boot, and then every 24 hours that the computer is running. This is consistent with the old disk space event, and is fine for a first pass. But ultimately we want to send this event at most every 24 hours, regardless of whether the computer is rebooted. From the Phabricator ticket:

> This event should be sent at most once every 24 hours, even if the computer is rebooted during that time. (This is because in metrics-v3 we can no longer deduplicate multiple events for the same machine by looking at machine ID – machine ID is not stored.)

We can tackle that in a later PR.

tests/test-hwinfo.c
starnight: The test cases have different idea and target.
* `test_get_cpu_info_for_current_system` checks the returned CPU information.
* `test_get_computer_hwinfo` checks the RAM, disk and CPU combination and they should be in order.

wjt: I suppose so, but the latter one does do all the checks that the former does (assert_cpu_info_for_current_system() + implicitly eins_hwinfo_get_cpu_info()) – there is no case where `test_get_cpu_info_for_current_system` would fail and `test_get_computer_hwinfo` would pass.

I'm OK with leaving both tests though. They do check slightly different things and are not redundant the other way around: `/hwinfo/cpu/current` could pass while `/hwinfo/computer/current` fails, which would help narrow down the failure.

src/eins-hwinfo.c
starnight: ok!

src/eins-hwinfo.c
starnight: I prefer the later one which simplifies things.

src/eins-hwinfo.c
starnight: ok

src/eins-hwinfo.c
starnight: sure

src/eins-hwinfo.c
starnight: ok!

src/eins-hwinfo.c
starnight: Fixed in the other way as mentioned above.

src/eins-hwinfo.c
starnight: I keep original `CPUINFO_TYPE_STRING` and add `CPUINFO_ARRAY_TYPE_STRING`.  `CPUINFO_TYPE_STRING` will be used to create new CPU information array in fixed version.

src/eins-hwinfo.c
starnight: ok.  But, fixed with a little bit different way for "the other option" mentioned above.

wjt (commented):
This commit has `From: dev <Jian-Hong Pan>` – is your Git `user.name` and `user.email` not set correctly?

src/eins-hwinfo.c
```diff
@@ -105,159 +115,70 @@ round_to_nearest (guint64 size,
...
   guint64 total = 0;
   guint64 used = 0;
-  guint64 available = 0;
+  guint64 free = 0;
+
+  if (diskspace == NULL)
+    {
+      g_error ("Wrong memory address to hold disk space information");
+      return FALSE;
+    }
```
wjt, line 143: If you want to abort the process in this case (which `g_error` will do), just use:

```suggestion
  g_assert (diskspace != NULL);
```

If you do not want to abort the process, use:

```
  g_return_val_if_fail (diskspace != NULL, FALSE);
```

src/eins-hwinfo.c
```diff
@@ -105,159 +115,70 @@ round_to_nearest (guint64 size,
...
-                                       error);
+                                       &error);
   if (info == NULL)
     {
-      /* TODO: g_file_peek_path() from GLib 2.56. */
-      g_autofree gchar *path = g_file_get_path (file);
-      g_prefix_error (error, "Couldn't get disk space for %s: ", path);
-      return NULL;
+      const gchar *path = g_file_peek_path (file);
+      g_prefix_error (&error, "Couldn't get disk space for %s: ", path);
```
wjt, line 158: If you're not going to return `error` to the caller there is no point in prefixing it – nothing will ever show the error.

```suggestion
      g_warning ("Couldn't get disk space for %s: %s", path, error->message);
```

wjt: Personally I would prefer the `GError **` out argument, which the caller can either supply (and log warning), or ignore.

wjt (changes requested):
I'll push a patch as the easiest way to make suggestions :)

starnight:
> This commit has `From: dev <Jian-Hong Pan>` – is your Git `user.name` and `user.email` not set correctly?

Oops!  I forgot to edit the user information after re-install the new image on the disk.

src/eins-hwinfo.c
starnight: Thanks for the hint.  I did not know that special behavior of [g_error()](https://developer.gnome.org/glib/stable/glib-Message-Logging.html#g-error) and thought it is just log level, until this time. :)

src/eins-hwinfo.c
starnight: ok

wjt (approved):

PR #95, opened 2021-08-04
hwinfo: Record hardware information every 24 hours

starnight:
Originally, eos-metrics-instrumentation records the hardware information
after system boots and 24 hours which is measured in monotonic time.
However, it may may or may not count time that the system is suspended.

This patch:
* saves the next record time (current real time + 24 hours) into storage
  when the information is recorded.
* checks how many time and wait for first record after each boot.
* records the hardware information every 24 hours after first record
  since boot. The 24 hours period is timed with CLOCK_BOOTTIME.

https://phabricator.endlessm.com/T32369

### DIFF
--- a/src/eins-boottime-source.c
+++ b/src/eins-boottime-source.c
@@ -1,0 +1,187 @@
+/* Copyright 2018–2021 Endless OS Foundation LLC */
+
+/* This file is part of eos-metrics-instrumentation.
+ *
+ * eos-metrics-instrumentation is free software: you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 2 of the License, or (at your
+ * option) any later version.
+ *
+ * eos-metrics-instrumentation is distributed in the hope that it will be
+ * useful, but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
+ * Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with eos-metrics-instrumentation.  If not, see
+ * <http://www.gnu.org/licenses/>.
+ */
+
+#include "eins-boottime-source.h"
+
+#include <errno.h>
+#include <inttypes.h>
+#include <sys/timerfd.h>
+#include <time.h>
+#include <unistd.h>
+#include <glib.h>
+#include <glib/gstdio.h>
+#include <gio/gio.h>
+
+typedef struct {
+  GSource parent;
+
+  int fd;
+  gpointer tag;
+} EinsBoottimeSource;
+
+static gboolean
+eins_boottime_check (GSource *source)
+{
+  EinsBoottimeSource *self = (EinsBoottimeSource *) source;
+
+  return g_source_query_unix_fd (source, self->tag) != 0;
+}
+
+static gboolean
+eins_boottime_source_dispatch (GSource    *source,
+                               GSourceFunc callback,
+                               gpointer    user_data)
+{
+  EinsBoottimeSource *self = (EinsBoottimeSource *) source;
+  uint64_t n_expirations = 0;
+
+  if (callback == NULL)
+    {
+      g_warning ("Boottime source dispatched without callback. "
+                 "You must call g_source_set_callback().");
+      return G_SOURCE_REMOVE;
+    }
+
+  /* Must read from the FD to reset its ready state. */
+  if (read (self->fd, &n_expirations, sizeof (n_expirations)) < 0)
+    {
+      g_warning ("read() failed for timerfd: %s",
+                 g_strerror (errno));
+      return G_SOURCE_REMOVE;
+    }
+
+  return callback (user_data);
+}
+
+static void
+eins_boottime_source_finalize (GSource *source)
+{
+  EinsBoottimeSource *self = (EinsBoottimeSource *) source;
+  g_autoptr(GError) local_error = NULL;
+
+  if (!g_close (self->fd, &local_error))
+    g_warning ("Failed to close timerfd: %s", local_error->message);
+  self->fd = -1;
+}
+
+static const GSourceFuncs eins_boottime_source_funcs = {
+  .check = eins_boottime_check,
+  .dispatch = eins_boottime_source_dispatch,
+  .finalize = eins_boottime_source_finalize,
+};
+
+/**
+ * eins_boottime_source_new_useconds:
+ * @interval_us: the timeout interval, in microseconds
+ * @error: return location for a #GError, or %NULL
+ *
+ * Like g_timeout_source_new(), but uses `CLOCK_BOOTTIME` to account for time
+ * when the system is suspended.
+ *
+ * If @interval_us is set to zero, the #GSource will be ready the next time it's
+ * checked.
+ *
+ * @error will be set to a #GIOError if, for example, the process runs out
+ * of file descriptors.
+ *
+ * Returns: (transfer full): a new `CLOCK_BOOTTIME` #GSource, or %NULL with
+ *   @error set
+ */
+static GSource *
+eins_boottime_source_new_useconds (guint64   interval_us,
+                                   GError  **error)
+{
+  g_autoptr(GSource) source = NULL;
+  EinsBoottimeSource *self = NULL;
+  int fd;
+  struct itimerspec its = {
+    .it_interval = {
+      .tv_sec = interval_us / G_USEC_PER_SEC,
+      .tv_nsec = (interval_us % G_USEC_PER_SEC) * 1000,
+    }
+  };
+
+  its.it_value = its.it_interval;
+
+  /* Set the GError if timerfd_create() fails because it could be e.g. ENFILE
+   * which we should handle gracefully */
+  fd = timerfd_create (CLOCK_BOOTTIME, TFD_CLOEXEC | TFD_NONBLOCK);
+  if (G_UNLIKELY (fd < 0))
+    {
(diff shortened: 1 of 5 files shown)


src/eins-hwinfo.c
```diff
@@ -103,6 +101,59 @@
...
+static gboolean
+get_next_record_time (guint *record, GError **error)
+{
+  gsize len = sizeof (guint);
+
+  return g_file_get_contents (RECORD_TIME_FILE_PATH,
+		              (gchar **)&record,
+			      &len,
+			      error);
+}
```
wjt, line 119: I would prefer for this file to be a `.ini`-style GKeyFile, like the existing `/var/cache/eos-metrics-instrumentation/persistent-tallies`. It's easier to inspect by hand and involves less scary casting of pointers to read.

src/eins-timer.c
```diff
@@ -0,0 +1,210 @@
+/* Copyright 2021 Endless OS Foundation LLC. */
```
wjt, line 1: I believe this code comes in part from eos-payg, so its copyright date should be earlier.

wjt: I prefer the naming `eins-boottime-source.[ch]`.

src/eins-timer.c
```diff
@@ -0,0 +1,210 @@
...
+  id = g_source_attach (source, NULL);
+  g_source_unref (source);
+
+  return id;
+}
+
+/**
+ * eins_get_realtime:
+ *
+ * Like g_get_real_time(), but the precision is only seconds.
```
wjt, line 192: This function returns `guint` which is a 4-byte type on x86-64. `struct timespec`'s `tv_sec` field is a `time_t` which is 8 bytes wide. https://en.wikipedia.org/wiki/Year_2038_problem .

 `g_get_real_time()` returns `gint64`. I suggest you just use `g_get_real_time()` at each call site, and either divide it by `G_USEC_PER_SEC` or just work in microseconds. I don't think there's much advantage of aligning to seconds.

(Using `g_timeout_add_seconds` rather than `g_timeout_add` allows GLib to merge nearby timers together to avoid unnecessary wakeups, but that is not relevant here – we are using a kernel-provided timerfd, and I can't see any way to tell the kernel "it doesn't need to be precise".)

src/eins-hwinfo.c
```diff
@@ -394,18 +459,33 @@ record_computer_hwinfo (gpointer unused)
...
 {
-  record_computer_hwinfo (NULL);
+  guint wait = get_wait_time_for_next_record ();
 
-  /* Record the computer hardware info every 24 hours. */
-  g_timeout_add_seconds (RECORD_COMPUTER_HWINFO_INTERVAL_SECONDS,
-                         record_computer_hwinfo,
-                         NULL);
+  /* If the wait time is less than 10mins, just skip the wait time. */
+  if (wait > 60u * 10u)
```
wjt, line 482: I don't think this special case is correct. The intention is that this event is recorded at most once every 24 hours. This allows us to slice the data into 24-hour chunks and be reasonably sure we are not double-counting any systems. If we record the event every 23h50m, that property of the data isn't true any more.

src/eins-hwinfo.c
```diff
@@ -394,18 +459,33 @@ record_computer_hwinfo (gpointer unused)
...
                                       COMPUTER_HWINFO_EVENT,
                                       g_steal_pointer (&payload));
+  set_next_record_time ();
+
+  /* The interval of first record after each boot usually is not 24 hour. */
+  if (record_first_time ())
+    {
+      eins_boottimeout_add_seconds (RECORD_COMPUTER_HWINFO_INTERVAL_SECONDS,
+                                    record_computer_hwinfo,
+                                    NULL);
```
wjt, line 469: You don't actually need this helper function with static variable – you can pass data to the callback. Suggested changes:

```suggestion
  if (GPOINTER_TO_UINT (is_first_call))
    {
      eins_boottimeout_add_seconds (RECORD_COMPUTER_HWINFO_INTERVAL_SECONDS,
                                    record_computer_hwinfo,
                                    GUINT_TO_POINTER (FALSE));
```

and others at the relevant points.

src/eins-hwinfo.c
```diff
@@ -385,6 +436,20 @@ eins_hwinfo_get_computer_hwinfo (void)
...
+    {
+      first = FALSE;
+      return TRUE;
+    }
+
+  return FALSE;
+}
+
 static gboolean
 record_computer_hwinfo (gpointer unused)
```
wjt, line 454:
```suggestion
record_computer_hwinfo (gpointer is_first_call)
```

src/eins-hwinfo.c
```diff
@@ -394,18 +459,33 @@ record_computer_hwinfo (gpointer unused)
...
-                         record_computer_hwinfo,
-                         NULL);
+  /* If the wait time is less than 10mins, just skip the wait time. */
+  if (wait > 60u * 10u)
+    {
+      eins_boottimeout_add_seconds (wait, record_computer_hwinfo, NULL);
+      return;
+    }
+
+  record_computer_hwinfo (NULL);
```
wjt, line 488:
```suggestion
    eins_boottimeout_add_seconds (wait, record_computer_hwinfo, GUINT_TO_POINTER (TRUE));
```

src/eins-hwinfo.c
wjt: Something like:

```
g_autoptr(GKeyFile) kf = g_key_file_new ();

g_key_file_load_from_file (kf, RECORD_TIME_FILE_PATH, G_KEY_FILE_NONE, NULL);
return g_key_file_get_int64 (kf, "hwinfo", "next-record-time", NULL);
```

src/eins-hwinfo.c
```diff
@@ -103,6 +101,59 @@
...
+{
+  guint now, next;
+  gssize len = sizeof (guint);
+  g_autoptr(GError) error = NULL;
+
+  now = eins_get_realtime ();
+  next = now + RECORD_COMPUTER_HWINFO_INTERVAL_SECONDS;
+
+  if (!g_file_set_contents (RECORD_TIME_FILE_PATH, (gchar *)&next, len, &error))
+    g_warning ("Failed to write " RECORD_TIME_FILE_PATH ": %s", error->message);
```
wjt, line 154: As above please use a GKeyFile.

src/eins-hwinfo.c
```diff
@@ -385,6 +436,20 @@ eins_hwinfo_get_computer_hwinfo (void)
...
+
+  if (first)
+    {
+      first = FALSE;
+      return TRUE;
+    }
+
+  return FALSE;
+}
+
```
wjt, line 452:
```suggestion
```

wjt (changes requested):
Nice spelunking through GLib timers and the one from eos-payg!

src/eins-hwinfo.c
starnight: The idea is the wait time before first record after each boot, not the normal period.  I think about the PAYG allowable time before entering new credit.  However, I understand people will reboot again and again, even the credit is timeout.  So, next boot will catch it.  Therefore, we can remove this special case.

src/eins-hwinfo.c
starnight: ok

src/eins-timer.c
starnight: It is due to the interval's type guint of `g_timeout_add_seconds()`.  Working in microseconds sounds like a good idea.

src/eins-hwinfo.c
starnight: ok

src/eins-hwinfo.c
starnight: ok

src/eins-timer.c
starnight: ok

src/eins-boottime-source.c
```diff
@@ -0,0 +1,187 @@
+/* Copyright 2018 Endless Mobile, Inc. */
```
wjt, line 1:
```suggestion
/* Copyright 2018–2021 Endless OS Foundation LLC */
```

src/eins-boottime-source.h
```diff
@@ -0,0 +1,24 @@
+/* Copyright 2018 Endless Mobile, Inc. */
```
wjt, line 1:
```suggestion
/* Copyright 2018–2021 Endless OS Foundation LLC */
```

src/eins-hwinfo.c
```diff
@@ -103,6 +101,57 @@
...
+/* Get wait time in microseconds */
+static guint64
+get_wait_time_for_next_record (void)
+{
+  guint64 now, record;
+
+  record = get_next_record_time ();
+  now = g_get_real_time ();
```
wjt, line 131: This returns `gint64`, which is signed, not `guint64`, which is not. Why not make get_next_record_time() return a `gint64` and correct the type of these variables?

It is true that in practice these values are always positive. No I don't know why it returns a signed value. :)

This function can still return `guint64`. We know that both branches below return a non-negative number.

wjt (changes requested):
Just a couple of minor changes!

src/eins-hwinfo.c
starnight: The datatype bothers me and makes me thinking these days.  The original idea was `time_t`, which is the best one.  However, `g_get_real_time()` return a value with datatype `gint64`.  So, I think where is the best point to unify the datatype as well.

src/eins-boottime-source.c
starnight: ok

src/eins-boottime-source.h
starnight: ok

wjt (approved):

PR #96, opened 2021-08-09
hwinfo: re-define RECORD_COMPUTER_HWINFO_INTERVAL_USECONDS

starnight:
Define RECORD_COMPUTER_HWINFO_INTERVAL_USECONDS as the constant, instead
of calculating formula to avoid unexpected compiling error.

https://phabricator.endlessm.com/T32369


src/eins-hwinfo.c
```diff
@@ -102,7 +102,7 @@
   ROOTFS_SPACE_TYPE_STRING "@" CPUINFO_ARRAY_TYPE_STRING ")"
 
 /* 24 hours */
-#define RECORD_COMPUTER_HWINFO_INTERVAL_USECONDS (G_USEC_PER_SEC * 60u * 60u * 24u)
+#define RECORD_COMPUTER_HWINFO_INTERVAL_USECONDS (86400000000u)
```
wjt, line 105: The commit message says "compiling error". Can you give more detail on why the old approach didn't work? Was it being performed with 32-bit arithmetic? Can the components of the multiplication be given some annotation to make them 64-bit literals?

starnight: Found constant `G_TIME_SPAN_DAY`!

wjt (approved):

PR #99, opened 2022-03-02
Miscellaneous cleanup

wjt:
Some of this is deleting code that has been either unused or used-but-vestigial since changes leading up to EOS 4.

The last few patches are warning fixes, with a confession that this is the prelude to a Meson port.


src/eos-metrics-instrumentation.c
```diff
@@ -290,18 +264,26 @@ quit_main_loop (GMainLoop *main_loop)
...
 {
+  g_autoptr(GOptionContext) context = NULL;
+  g_autoptr(GError) error = NULL;
+
+  context = g_option_context_new ("- record metrics for systemwide events");
+  if (!g_option_context_parse (context, &argc, &argv, &error))
+    {
+      g_print ("option parsing failed: %s\n", error->message);
+      exit (1);
+      }
```
dbnicholson, line 278: Closing brace indentation and I think `g_printerr` rather than `g_print`.

dbnicholson (changes requested):

src/eos-metrics-instrumentation.c
wjt: I copy-pasted this from the GLib documentation :-)

wjt:
This is obviously not ready for review. I'm sure I built it locally but apparently not…

tools/Makefile.am.inc
```diff
@@ -16,21 +16,6 @@
...
-	$(NULL)
-
-tools_eos_network_id_CFLAGS = \
-	$(EOS_INSTRUMENTATION_CFLAGS) \
-	-I$(top_srcdir)/src \
-	$(NULL)
-
-tools_eos_network_id_LDADD = $(EOS_INSTRUMENTATION_LIBS)
-
 nodist_bin_SCRIPTS = tools/eos-label-location
```
wjt, line 19: The problem appears to be that Automake is not making the `tools` subdirectory in the builddir. I had not tested an out-of-tree build.

My autotools-ese is too rusty to understand what can be done to cause automake to make the directory. I'll look more later.

wjt: Ugh, maybe I should just chuck the meson port onto this branch and not think about this.

wjt: Added a fix which is gross but does work.

dbnicholson: I think that's the right thing to do. You can the same in `data/Makefile.am.inc`.

dbnicholson (approved):

PR #100, opened 2022-03-04
Port to Meson

wjt:
Packaging in #102


tests/meson.build
```diff
@@ -0,0 +1,19 @@
+test_sources = {
+    'test-hwinfo': [
+        '../src/eins-hwinfo.c',
+        '../src/eins-hwinfo.h',
+        '../src/eins-boottime-source.c',
+        '../src/eins-boottime-source.h',
```
eli-schwartz, line 6: Maybe you should build these 2 files once in the src/ directory as an uninstalled `static_library()`, then link it into both the `eos-metrics-instrumentation` program and this test program?

wjt: Thanks, I've made that change.

wjt:
retest this please

src/meson.build
```diff
@@ -0,0 +1,37 @@
...
+    ],
+    sources: [
+        'eos-metrics-instrumentation.c',
+    ],
+    install: true,
+    install_dir: libexec_dir,
+)
+
+crash_metrics = executable('eos-crash-metrics',
+    # TODO: too much
```
pwithnall, line 30: Is this TODO comment something you plan to address before merging?

meson.build
```diff
@@ -0,0 +1,46 @@
+
```
pwithnall, line 1: Leading blank line could be removed

configure.ac
```diff
@@ -1,180 +0,0 @@
...
-EOS_INSTRUMENTATION_VERSION=_EOS_INSTRUMENTATION_VERSION_MACRO
-AC_SUBST(EOS_INSTRUMENTATION_VERSION)
-EOS_INSTRUMENTATION_MAJOR_VERSION=_EOS_INSTRUMENTATION_MAJOR_VERSION_MACRO
-AC_SUBST(EOS_INSTRUMENTATION_MAJOR_VERSION)
-AC_DEFINE([EOS_INSTRUMENTATION_MAJOR_VERSION], [_EOS_INSTRUMENTATION_MAJOR_VERSION_MACRO], [Major version])
-EOS_INSTRUMENTATION_MINOR_VERSION=_EOS_INSTRUMENTATION_MINOR_VERSION_MACRO
-EOS_INSTRUMENTATION_MICRO_VERSION=_EOS_INSTRUMENTATION_MICRO_VERSION_MACRO
-AC_DEFINE([EOS_INSTRUMENTATION_MINOR_VERSION], [_EOS_INSTRUMENTATION_MINOR_VERSION_MACRO], [Minor version])
-AC_DEFINE([EOS_INSTRUMENTATION_MICRO_VERSION], [_EOS_INSTRUMENTATION_MICRO_VERSION_MACRO], [Micro version])
-AC_DEFINE([EOS_INSTRUMENTATION_VERSION], [_EOS_INSTRUMENTATION_VERSION_MACRO], [Full version number])
```
pwithnall, line 81: Note for posterity: these macros have all been dropped in the Meson version, presumably because they aren’t used anywhere in the code and the project installs no public headers.

pwithnall (approved):
LGTM. Two mostly irrelevant comments

configure.ac
wjt: Yes, this is not a library, it's a pair of executables.

src/meson.build
wjt: Good catch, I missed this!

pwithnall (approved):
LGTM

PR #101, opened 2022-03-04
Use start_aggregate_timer_with_uid() for session time

wjt:
Previously, the daemon would ascribe the UID of the calling process, in
this case 0 (root), to the event when aggregating. This new API allows
the caller to override this UID, removing the need for an aggregate key.

https://phabricator.endlessm.com/T33252

Depends on https://github.com/endlessm/eos-metrics/pull/47.


dbnicholson (commented):
Looks good. I'll kick the CI again after merging the eos-metrics PR.

dbnicholson:
retest this please

dbnicholson (approved):

PR #102, opened 2022-03-07
Build-depend on Meson

wjt:
Counterpart to #100. One of these days we should combine these brancehs.


pwithnall (approved):
:ship: :rocket: :ship:

PR #113, opened 2024-03-24
Report CPU flags in hwinfo

wjt:
Draft: Need to revise the server to accept a different type.


src/eins-hwinfo.c
```diff
@@ -232,7 +233,7 @@ typedef struct _LscpuFieldType {
    * field names, which is presumably a bug caused by re-using the field names
    * from the colon-separated default output.
    */
-  const gchar *names[3];
+  const gchar *names[4];
```
wjt, line 236: Not needed, this array size is the number of different names by which the field can be known, not the length of `LSCPU_FIELDS`

PR #114, opened 2024-05-02
debian: Build-depend on systemd-dev

wjt:
systemd.pc, which is used to look up the paths to install a systemd unit, tmpfiles.d snippet and sysctl.d snippet, is part of this package.

When we updated systemd in EOS 6, we dropped our patch for the the `systemd` binary package in previous branches which marked it as `Priority: required`, instead leaving it as `Priority: important`. When it was `required`, it would always be present in the buildroot, and so would `systemd-dev` which it (surprisingly!) depends on; but now it is not. 

Explicitly declare this build dependency, as we should have been doing all along.

Also add a `.gitignore`. My bad for removing it in 6eb0bbd but in my defence the version I removed didn't ignore these files.

https://phabricator.endlessm.com/T35070#990194
https://phabricator.endlessm.com/T35352

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -13,6 +13,7 @@ Build-Depends:
 	meson,
 	python3-dbus,
 	python3-dbusmock,
+	systemd-dev,
 	util-linux (>= 2.32),
 
 Package: eos-metrics-instrumentation
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,5 @@
+/debian/.debhelper/
+/debian/debhelper-build-stamp
+/debian/eos-metrics-instrumentation.substvars
+/debian/eos-metrics-instrumentation/
+/debian/files


debian/control
```diff
@@ -10,6 +10,7 @@ Build-Depends:
 	libgtop2-dev,
 	libjson-glib-dev,
 	libostree-dev,
+	libsystemd-dev,
```
starnight, line 13: I think here is `systemd-dev` to fix the build failure:
```
Run-time dependency systemd found: NO (tried pkgconfig)

../../meson.build:31:0: ERROR: Dependency "systemd" not found, tried pkgconfig
```

starnight (changes requested):

debian/control
wjt: Quite right.

wjt: I even got it right in the commit message! 😁

starnight (approved):
Thanks!

